PR #1, opened 2018-08-15
Forms

JoeyFaulkner:


examples/formbot/domain.yml
```diff
@@ -0,0 +1,66 @@
...
+- utter_explain_details_restaurant
+- utter_suggest_restaurant
+- utter_ask_cuisine
+- utter_ask_people
+- utter_ask_location
+- utter_ask_price
+- utter_book_restaurant
+- form_action
+
+form_start_actions:
```
tmbo: what do we need this for? isn't it enough that that action emits the start_form event?

examples/formbot/myactions.py
```diff
@@ -0,0 +1,75 @@
...
+        form = executor.forms[tracker.active_form]
+        still_to_go = form.check_unfilled_slots(tracker)
+        complete = len(still_to_go) == 0
+        return [SlotSet('form_complete', complete), EndForm()]
+
+
+class RestaurantForm(SimpleForm):
+    def __init__(self):
+        name = 'restaurant_form'
+        slot_dict = {
```
tmbo: i'd like to avoid calling the attributes, `xxx_dict` - how about `fields`, `breakout_intents`, ... or something that better describes what these attributes configure

examples/formbot/myactions.py
```diff
@@ -0,0 +1,75 @@
...
+        finish_action = "end_form"
+
+        exit_dict = {
+            "goodbye": "end_form",
+            "request_hotel": "end_form"
+        }
+
+        chitchat_dict = {"chitchat": "utter_chitchat"}
+
+        details_intent = "utter_ask_details"
```
tmbo: is this an intent or a bot utterance?

examples/formbot/myactions.py
```diff
@@ -0,0 +1,75 @@
...
+    def run(self, dispatcher, tracker, domain, executor):
+        form = executor.forms[tracker.active_form]
+        still_to_go = form.check_unfilled_slots(tracker)
+        complete = len(still_to_go) == 0
+        return [SlotSet('form_complete', complete), EndForm()]
+
+
+class RestaurantForm(SimpleForm):
+    def __init__(self):
+        name = 'restaurant_form'
```
tmbo: all of these are static values, why not make them class attributes?

rasa_core_sdk/events.py
```diff
@@ -126,3 +126,28 @@ def AgentUttered(text=None, data=None, timestamp=None):
...
+# noinspection PyPep8Naming
+def EndForm(timestamp=None):
+    return {
+        "event": "end_form",
+        "timestamp": timestamp
+    }
+
+
+# noinspection PyPep8Naming
+def FormListen(timestamp=None):
```
tmbo: what do we need this for?

tests/test_form.py
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        failure_action = 'utter_human_hand_off'
+        super(TestForm, self).__init__(name, slot_dict,
+                                       finish_action, exit_dict,
+                                       chitchat_dict, details_intent,
+                                       rules, failure_action=failure_action)
+
+
+def test_endpoint():
```
tmbo: ?

examples/formbot/myactions.py
JoeyFaulkner: its an intent, but maybe that is confusing and i should rename it in domain

examples/formbot/domain.yml
JoeyFaulkner: they're needed here https://github.com/RasaHQ/rasa_core/pull/878/files#diff-044625afbb4ad4ce7b830a6cd066196aR332 because during training the actions aren't accessible or carried out and message/form_actions need to be added so that the memoization policy works

rasa_core_sdk/events.py
JoeyFaulkner: see https://github.com/RasaHQ/rasa_core/pull/878#discussion_r210857097

rasa_core_sdk/forms.py
```diff
@@ -81,82 +82,179 @@ def extract(self, tracker):
...
-    def get_requested_slot(self, tracker):
-        requested_slot = tracker.get_slot(FORM_SLOT_NAME)
-
-        required = self.required_fields()
-
-        if self.RANDOMIZE:
-            random.shuffle(required)
-
-        if requested_slot is None:
+    def run(self, dispatcher, tracker, domain, executor):
```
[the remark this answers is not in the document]
JoeyFaulkner: this way we can have a single action which covers multiple different forms (whichever is tracker.active_form)

examples/formbot/myactions.py
JoeyFaulkner: they are validated and set as attribjutes in the __init__ for SimpleForm

rasa_core_sdk/events.py
tmbo: I am not sure we added this yet, but can we not use a followup action? forcing the bot to run action listen as the next action?

JoeyFaulkner: it hasnt been added yet, i use the follow up action in core but not on the SDK

tmbo:
Outdated, we are working on a new approach for this.

PR #12, opened 2018-10-03
separate logic for which slot to request next 

amn41:
users may wish to override this logic (e.g. if there is conditional logic, only request slot B if slot A has value X)

cc @MetcalfeTom

### DIFF
--- a/rasa_core_sdk/forms.py
+++ b/rasa_core_sdk/forms.py
@@ -111,6 +111,12 @@ class FormAction(Action):
 
         return [Form(None), SlotSet(REQUESTED_SLOT, None)]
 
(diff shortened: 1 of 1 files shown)


MetcalfeTom:
I like the concept of separating it, however I don't think it fully solves the problem because the utterance template may also differ depending on the values of the previous slots.  In this case `FormAction.run()` would have to check the slot value twice - could we move `dispatcher.utter_template()` into `next_slot_to_request()`?

Ghostvv:
@MetcalfeTom I think all FormAction.run() helper methods should have the same signature as run, so moving `dispatcher.utter_template()` into `next_slot_to_request(dispatcher, tracker, domain)` makes sense

PR #18, opened 2018-10-16
formatting --actions arg so that i work both in the directory style a…

### LINKED ISSUE #16: throw error if actions module is passed as directory rather than module
`--actions directory.actions` vs `--actions directory/actions`

EPedrotti:
…nd module style, fixes #16

### DIFF
--- a/rasa_core_sdk/endpoint.py
+++ b/rasa_core_sdk/endpoint.py
@@ -9,7 +9,6 @@ import logging
 from flask import Flask, jsonify, request
 from flask_cors import CORS, cross_origin
 from gevent.pywsgi import WSGIServer
-
 from rasa_core_sdk.executor import ActionExecutor
 
 DEFAULT_SERVER_PORT = 5055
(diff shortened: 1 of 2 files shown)


rasa_core_sdk/endpoint.py
```diff
@@ -9,35 +9,41 @@
...
 from gevent.pywsgi import WSGIServer
-
 from rasa_core_sdk.executor import ActionExecutor
 
 DEFAULT_SERVER_PORT = 5055
 
 logger = logging.getLogger(__name__)
 
 
+def action_arg(action):
```
tmbo: I am split on this: should we do this or rather reject anything that contains `/`? The reason I am worried is, that the path components might not actually be proper python packages (e.g. missing `__init__.py`) - so the loading with the converted name will fail as well

EPedrotti: I think the missing __init__.py can happen I both cases, with the `/` or with the `.`. 
But if there are other reasons i can just reject anything with `/`, just let me know.

tmbo: yes it can definitely happen in both cases, but if the user provides it with `.` we can assume he knows it is supposed to be a python package name, if he provides it with `/` he might not know it should be a python package and the error message he will get will not make sense to him

rasa_core_sdk/endpoint.py
```diff
@@ -9,35 +9,43 @@
...
+    if "/" in action:
+        raise argparse.ArgumentTypeError(
+            'Invalid actions format. Actions file should be a python module and passed with module notation (e.g. '
+            'directory.actions).')
```
akelad: Ok I know this was already merged, but i'd like this error message to take the argument they passed, e.g. `bot/actions` and say something like "Did you mean bot.actions?"

Issue #19, opened 2018-10-17
Create rasa_core_sdk docker container


wochinge:
There is currently no rasa_core_sdk docker container which can be used to run an action endpoint.
This image is also required to build the suggested docker compose file of issue https://github.com/RasaHQ/rasa_core/issues/845.

Steps todo:
* create docker image
* provide information on how to enhance it with custom dependencies
* automate build and pushing of the image

PR #20, opened 2018-10-17
#19 - added docker image

wochinge:
**Motivation**:
Build a lean standalone Docker image for the rasa core sdk which can be 
* enhanced easily
* connected to a rasa core contaner
* required to implement https://github.com/RasaHQ/rasa_core/issues/845

**Proposed changes**:
- adding Dockerfile including an entrypoint file for simpler execution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7-slim
+FROM python:3.6-slim
 
 SHELL ["/bin/bash", "-c"]
 
@@ -12,11 +12,11 @@ RUN apt-get update -qq && \
 WORKDIR /app
 
 # Copy as early as possible so we can cache ...
-ADD requirements.txt .
+COPY requirements.txt .
 
 RUN pip install -r requirements.txt --no-cache-dir
 
-ADD . .
+COPY . .
 
 RUN pip install -e . --no-cache-dir
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM python:2.7-slim
+
+SHELL ["/bin/bash", "-c"]
+
+RUN apt-get update -qq && \
+  apt-get install -y --no-install-recommends \
+  build-essential && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+  mkdir /app
+
+WORKDIR /app
+
+# Copy as early as possible so we can cache ...
+ADD requirements.txt .
+
+RUN pip install -r requirements.txt --no-cache-dir
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM python:2.7-slim
```
tmbo: should be 3.6 - I think we shouldn't use 2 in any of our docker files

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  apt-get install -y --no-install-recommends \
+  build-essential && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+  mkdir /app
+
+WORKDIR /app
+
+# Copy as early as possible so we can cache ...
+ADD requirements.txt .
```
tmbo: i think `COPY` is the better choice here (as we don't need fancy `ADD` features)

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN pip install -r requirements.txt --no-cache-dir
+
+ADD . .
```
tmbo: `COPY`

entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
```
tmbo: let's be even more strict and use `set -Eeuo pipefail`

entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+    start)
+        exec python -m rasa_core_sdk.endpoint "${@:2}"
+        ;;
+    run)
+        exec "${@:2}"
+        ;;
+    *)
+        print_help
+        ;;
+esac
```
tmbo: newline

README.md
```diff
@@ -24,7 +24,41 @@ pip install rasa_core_sdk
...
+Before starting the action server ensure that the folder containing your actions is handled as Python module and therefore has to contain a file called `__init__.py`
+
+Then start the action server using:
+
+```bash
+docker run --mount type=bind,source=<ABSOLUTE_PATH_TO_YOUR_ACTIONS>,target=/app/actions \
+	rasa/rasa_core_sdk:latest 
+```
+
+The action server is then avaible at `http://localhost:5055/webhook`.
```
tmbo: aren't we missing a port mapping from inside the container to the host machine? otherwise this webhook can't be reached

tmbo:
once merged, this should be build on docker hub

README.md
wochinge: added port mapping

PR #28, opened 2018-11-27
Validate methods for slots

ksurya:
**Proposed changes**:
-  `validate()` function calls optionally `validate_{slot_name}` functions to validate slot values
- `validate_{slot_name}` returns None for invalid values

- [ ] updated the changelog

### DIFF
--- a/rasa_core_sdk/forms.py
+++ b/rasa_core_sdk/forms.py
@@ -251,8 +251,14 @@ class FormAction(Action):
         # extract requested slot
         slot_to_fill = tracker.get_slot(REQUESTED_SLOT)
         if slot_to_fill:
-            slot_values.update(self.extract_requested_slot(dispatcher,
-                                                           tracker, domain))
+            for slot, value in self.extract_requested_slot(dispatcher,
+                                                           tracker,
+                                                           domain).items():
+                validate_func = getattr(self, "validate_{}".format(slot),
+                                        lambda *x: value)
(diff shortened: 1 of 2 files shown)


rasa_core_sdk/forms.py
```diff
@@ -251,8 +251,16 @@ def validate(self, dispatcher, tracker, domain):
...
-                                                           tracker, domain))
+            for slot, value in self.extract_requested_slot(dispatcher, tracker,
+                                                           domain).items():
+                validate_func = getattr(self, "validate_{}".format(slot),
+                                        lambda *x: value)
+                value = validate_func(value, dispatcher, tracker, domain)
+                if value:
+                    slot_values[slot] = value
+                else:
+                    tracker.slots[slot] = None
```
Ghostvv: it will not override slot value in `rasa_core`, in order to set the value to `None`, and action should return `SlotSet(slot, None)` event, so I think it makes sense to just write `slot_values[slot] = value` without `if`

ksurya:
Thank you for the feedback, I've made the requested change. Would you also want changelog in Rasa Core updated?

Ghostvv:
thanks for updating. No we don't need to add to `core` change log, but we need to update `core` docs to explain how to use it, and that these functions should return `None` if validation failed.
It would be cool also to update `formbot` example in `core` to use these helper functions

Ghostvv:
sorry, didn't see this PR: https://github.com/RasaHQ/rasa_core/pull/1422#pullrequestreview-180980234

PR #40, opened 2019-01-31
Added coloredlogs to the SDK

Patrick-DS:
**Proposed changes**:
- *Added coloredlogs to the SDK.* I assume I just had to add it where the \_\_main__ gets called, since all the files are getting their logger from the line 'logger = logging.getLogger(\_\_name__)" so I just imported the utils from rasa and added the same lines that are in the rasa_core library. It worked for me.

- [ ] updated the changelog

### DIFF
--- a/rasa_core_sdk/endpoint.py
+++ b/rasa_core_sdk/endpoint.py
@@ -15,6 +15,8 @@ from rasa_core_sdk.executor import ActionExecutor
 from rasa_core_sdk import ActionExecutionRejection
 import rasa_core_sdk
 
+from rasa_core_sdk import utils
+
 DEFAULT_SERVER_PORT = 5055
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


Patrick-DS:
Actually it's not literally a dependency on rasa_core, rasa_core_sdk has the exact same utils file containing the exact same command. I just need to pull the function from the right import statement and it'll work the same.

akelad:
well, it is a dependency on rasa_core because you're importing from rasa_core :)

akelad:
Ah wait got it, yes just modify it to import from rasa_core_sdk instead

Patrick-DS:
Since my first commit was not quite correct, I did another one to add what was necessary and squashed the two commits (because it really is just one commit, I won't leave my mistakes in there). Do I need anything to ask for the PR to be reviewed again, or can you just approve it now?

I had to add the function I added in the utils of the SDK because that function was only in the rasa_core utils file and I needed it to add log levels, which are required by configure_colored_logging.

PR #46, opened 2019-02-06
added latest_input_channel to tracker

smn-snkl:
I just realized that the `tracker.get_latest_input_channel` feature I've had in my fork for a while is not part of the official `rasa_core_sdk` yet (but is part of `rasa_core`). Because there's been multiple requests about this on the forum (e.g. [here](https://forum.rasa.com/t/how-to-detect-chat-platform-in-actions-py/2991/8) or [here](https://forum.rasa.com/t/action-latest-action-name-followup-action-latest-input-channel-return-none/4157)) I wanted to push this change.

**Proposed changes**:
- Similar to `rasa_core`'s approach, we reverse traverse tracker.events until latest `user` event is found. Then return `event.input_channel` 

- [x] updated the changelog


MetcalfeTom:
Thanks for the help @smn-snkl, I'll merge this into the code once the tests pass

(Regarding the pep8 fix, just wanted to keep the code in-line with our style guide)

smn-snkl:
@MetcalfeTom cool, thanks! Just saw that the check was failing because of an style issue and quickly made it pass. All good.

PR #47, opened 2019-02-06
Allow for early form deactivation during validation

MetcalfeTom:
If a slot could not be validated properly, we may want to deactivate the form immediately (e.g. without submitting).  Currently, adding `self._deactivate()` to the events would still cause the rest of the `run()` method to continue.

**Proposed changes**:
- Add a check in the `run()` method that we didn't already try to deactivate
- Rename method to `deativate()` (i.e. remove the "protected" status)

- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,13 +15,14 @@ Added
 -----
 - add optional `validate_{slot}` methods to `FormAction`
 - forms can now be deactivated during the validation function by returning
-  `self._deactivate()`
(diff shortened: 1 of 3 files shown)


MetcalfeTom:
@Ghostvv only mention of deactivating a form in `rasa_core` is [here](https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/actions/action.py#L213) (luckily we don't use the method, so no change necessary)

Will add a docs update now in that repo

PR #49, opened 2019-02-15
Abstract Action Detection

smn-snkl:
As discussed [here](https://github.com/RasaHQ/rasa_core_sdk/issues/29), I suggest addeding a "abstract action detection" mechanism to the executor's `register_package` to allow subclassing Action or FormAction to create customized behavior.

**Proposed changes**:
- extract abstract value from each action's `Meta` class. Defaults to `False`, thereby not affecting current behavior

Example is discussed [here](https://github.com/RasaHQ/rasa_core_sdk/issues/29#issuecomment-463959719)

- [x] updated the changelog


smn-snkl:
@tmbo : Feature is tested and ready to be reviewed. Let me know what you think

tmbo:
@smn-snkl Just thought about this again, and I think we should make that a little simpler to use for the end user. Your dict access approach brought me to an idea, can we use something like this?

```python
def rasa_ignore(original_class):
    original_class.is_action_class_ignored = True
    return original_class

@rasa_ignore
class A:
    pass

class B(A):
    pass


b = B()
a = A()

assert not B.__dict__.get("is_action_class_ignored", False)
assert A.__dict__.get("is_action_class_ignored")
```

smn-snkl:
@tmbo : I'm not sure if that's actually easier. My approach is pretty much the same, except that you just set the `ignore` flag on the base action directly like this:

```
class A:
    class Meta:
        abstract=True

class B(A):
    pass
```
We could also just do:
```
class A:
    abstract=True

class B(A):
    pass
```
Which would now be even easier for the user since he doesn't even need to import anything. However, I'd rather keep the `class Meta` to protect the user from overriding the `abstract` class variable by accident.

As mentioned above (and learned from the Django project) having this Meta class object is also very nice for other customizations, e.g. I myself customized FormAction to force asking all slots, **even already filled ones**, I just needed to add:
```
class A:
    class Meta:
        abstract=True
        force_slots=True
```

tmbo:
Ok I like the Meta class idea let's do that :+1:

adamwojt:
Would be great if this was documented. It's quite important stuff and I have to follow comments from PR and there is only small thing in Changelog :)
Also following tutorial from PR, I still get:
```
    raise NotImplementedError("An action must implement a name")
NotImplementedError: An action must implement a name
```
And if name is provided then action is registered despite adding Meta stuff. Any solutions?

PR #55, opened 2019-02-27
Validate for all extracted slots

yobekiko:
**Proposed changes**:
- Modified `validate` function to validate all extracted slots. Previously it validates only requested slot but not extracted other slots.

- [ ] updated the changelog

### DIFF
--- a/rasa_core_sdk/forms.py
+++ b/rasa_core_sdk/forms.py
@@ -254,21 +254,22 @@ class FormAction(Action):
             slot_values.update(self.extract_requested_slot(dispatcher,
                                                            tracker,
                                                            domain))
-            for slot, value in slot_values.items():
-                validate_func = getattr(self, "validate_{}".format(slot),
-                                        lambda *x: value)
-                slot_values[slot] = validate_func(value, dispatcher, tracker,
-                                                  domain)
 
             if not slot_values:
                 # reject to execute the form action
                 # if some slot was requested but nothing was extracted
                 # it will allow other policies to predict another action
                 raise ActionExecutionRejection(self.name(),
-                                               "Failed to validate slot {0} "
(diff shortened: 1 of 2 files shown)


akelad:

rasa_core_sdk/forms.py
```diff
@@ -251,9 +251,10 @@ def validate(self, dispatcher, tracker, domain):
         # extract requested slot
         slot_to_fill = tracker.get_slot(REQUESTED_SLOT)
         if slot_to_fill:
-            for slot, value in self.extract_requested_slot(dispatcher,
+            slot_values.update(self.extract_requested_slot(dispatcher,
                                                            tracker,
-                                                           domain).items():
+                                                           domain))
+            for slot, value in slot_values.items():
```
Ghostvv: let's raise exception before validating, and rename `Failed to extract slot ...`, so that the form stop execution if nothing was extracted, but if something was extracted, but validation failed it will ask a question again

yobekiko: Do you mean before each slot (no matter requested or others) is validated, raise exception if the value is None? Or just check for requested slot?

Ghostvv: I suggest move the below code block:
```
            if not slot_values:
                # reject to execute the form action
                # if some slot was requested but nothing was extracted
                # it will allow other policies to predict another action
                raise ActionExecutionRejection(self.name(),
                                               "Failed to validate slot {0} "
                                               "with action {1}"
                                               "".format(slot_to_fill,
                                                         self.name()))
```
before line `257` and change `Failed to validate slot` to `Failed to extract slot`

yobekiko: Got it. That makes sense

yobekiko:
Done.

PR #63, opened 2019-04-04
make validate_{} functions return {slot: value} dicts instead of just values

erohmensing:
**Proposed changes**:
- validate_{} functions now return {slot: value} dicts instead of just value, so that they can update other slots as well 
- allows setting of other slots from within validate helper methods

- [x] updated the changelog


rasa_core_sdk/forms.py
```diff
@@ -311,16 +311,32 @@ def extract_requested_slot(
                     )
                     return {slot_to_fill: value}
 
-        logger.debug("Failed to extract requested slot '{}'" "".format(slot_to_fill))
+        logger.debug("Failed to extract requested slot '{}'".format(slot_to_fill))
         return {}
 
+    # def validate_<slot_name>(self, value, dispatcher, tracker, domain):
```
Ghostvv: @tmbo we want to provide generic example with all the in/out types here (like a template method), but users cannot directly override it in subclass, since the name will be different. Is it a good way to showcase it?

rasa_core_sdk/forms.py
```diff
@@ -343,9 +359,9 @@ def validate(self, dispatcher, tracker, domain):
 
             for slot, value in slot_values.items():
```
Ghostvv: @erohmensing this has to go outside `if slot_to_fill:`

erohmensing: @Ghostvv this is more for fixing the other issue (slots not getting validated if they were picked up by the input) right? Do you reckon we should fix that issue in a separate non-breaking change?

erohmensing: I'm still not convinced this is the entire fix because of [this line](https://github.com/RasaHQ/rasa_core_sdk/blob/fe65450190e05f3066e10bb37c689d8841a9ca3f/rasa_core_sdk/forms.py#L446) -- for me it wasn't validating if you put anything between the user input and the form, such as something like
```
*subscribe_newsletter{"email": "ella@ella.ella"}
    - utter_can_do
    - newsletter_form
```
then the latest action was `utter_can_do.` Should it not be allowed to put an utterance in there? cause I think it's a relatively common thing to write in a story

Ghostvv: oh, ye I forgot that this is `sdk` breaking changes.
Could you please create a separate PR
Regarding utterance before, I'm not sure here, since intent is no more "trigger" of the form, but rather `utter_can_do` is a trigger. I would create another helper method (the opposite of submit) that would be called during init and can be subclassed to put a first utterance there - this can be done in this PR

erohmensing: @akelad what do you think about another helper method for actions to be done when opening the form? I think this has to do with the comment you made on my sara PR about the fact that the `newsletter_form` should be right after the `subscribe_newsletter` intent. Does vova's idea make sense to you?

rasa_core_sdk/forms.py
tmbo: not sure I fully understand, so it would be in here as a comment?

erohmensing: yeah, I think that's the idea (not sure if you noticed but there's a whole function here, not just the first line). Another idea is we could write a full (non-commented) function for something titled `validate_example_slot()` and then the description would be something like "an example custom validation function for validating a slot named "example_slot". You can use this template to create custom validation functions for your slots", or something like that

tmbo: Ok in that case, how about the following: use the goolge python style `Example` section in the `validate` methods description to show how it should be used (this will make sure the syntax is correct, while this comment won't)

Ghostvv:
could you please add debug statement that all required slots are filled (maybe with printing values, similarly how we print slot values in core) before calling `submit` method

erohmensing:
@Ghostvv yes i can add that, will add to #67 instead since that will be in the next minor and this one's breaking

Ghostvv:
sure, sorry I put in wrong place then, too many similar PRs 🙈

rasa_core_sdk/forms.py
erohmensing: @Ghostvv i talked to tom, we're going to put the example in the docs instead

Ghostvv: sounds good

rasa_core_sdk/forms.py
```diff
@@ -339,9 +340,16 @@ def validate(self, dispatcher, tracker, domain):
                     "".format(slot_to_fill, self.name()),
                 )
 
-        for slot, value in slot_values.items():
+        for slot, value in list(slot_values.items()):
             validate_func = getattr(self, "validate_{}".format(slot), lambda *x: value)
```
Ghostvv: I guess this should be:
```suggestion
            validate_func = getattr(self, "validate_{}".format(slot), lambda *x: {slot: value})
```

erohmensing: oh yes, we totally changed this before and i missed it when i merged master, thanks for the catch!

erohmensing: now i'm thinking `{'slot_name': 'value'}` in my warning should actually be ` {'slot_name': value}`, since values don't have to be strings, right?

Ghostvv: I'd say {'{slot_name}': value}`?

PR #67, opened 2019-04-10
Validate prefilled slots

erohmensing:
**Proposed changes**:
- validate slots filled before start of form 
- debug logging for values of required slots before form submission
- debug logging for submission of form 

- [x] updated the changelog


erohmensing:
Screen Shot 2019-04-11 at 15 17 26
here's what the logging output will look like

rasa_core_sdk/forms.py
```diff
@@ -435,6 +446,24 @@ def _activate_if_required(self, tracker):
             logger.debug("Activated the form '{}'".format(self.name()))
             return [Form(self.name())]
 
+    def _validate_prefilled_slots(self, dispatcher, tracker, domain):
+        # type: (Tracker) -> List[Dict]
+        """Upon form activation, validate all required slots that are already
+           filled."""
```
Ghostvv: should be blank line after method description. Also could you create one-liner description and the rest should go after blank line. `"""` in the end on the new line for multi-line comment, e.g.:
```
    @staticmethod
    def required_slots(tracker):
        # type: (Tracker) -> List[Text]
        """A list of required slots that the form has to fill.

        Use `tracker` to request different list of slots
        depending on the state of the dialogue
        """

        raise NotImplementedError(
            "A form must implement required slots " "that it has to fill"
        )
```

Ghostvv: Looks like this method duplicates `validate`, can we simply call `validate` upon activation, e.g. `_activate_if_required(...)` method

rasa_core_sdk/forms.py
```diff
@@ -443,7 +472,11 @@ def _validate_if_required(self, dispatcher, tracker, domain):
             - the form is called after `action_listen`
             - form validation was not cancelled
         """
-        if tracker.latest_action_name == "action_listen" and tracker.active_form.get(
+        if tracker.latest_action_name == "form" and tracker.active_form.get(
+            "validate", True
+        ):
```
Ghostvv: this if looks awful, is it black formatting?

rasa_core_sdk/forms.py
```diff
@@ -443,7 +472,11 @@ def _validate_if_required(self, dispatcher, tracker, domain):
...
+            return self._validate_prefilled_slots(dispatcher, tracker, domain)
+        elif tracker.latest_action_name == "action_listen" and tracker.active_form.get(
```
Ghostvv: this if looks awful, is it black formatting?

rasa_core_sdk/forms.py
Ghostvv: Did you test it? I doubt `tracker.latest_action_name == "form"` is correct

rasa_core_sdk/forms.py
Ghostvv: I see, `validate` contains extraction. Then vise versa. let's call this from inside `validate`, renaming this method and passing slot names to be validated

Ghostvv: what do you think about introducing breaking changes and create separate `extract` method outside `validate`?

tests/test_forms.py
```diff
@@ -652,6 +652,31 @@ def required_slots(_tracker):
...
+            return ["some_slot", "some_other_slot"]
+
+        def validate_some_slot(self, value, dispatcher, tracker, domain):
+            if value == "some_value":
+                return "validated_value"
+
+    form = CustomFormAction()
+
+    tracker = Tracker(
+        "default", {"some_slot": "some_value"}, {}, [], False, None, {}, "form"
```
Ghostvv: I think this test overfits to your code. Hardcoding concrete previous action name `form` and then testing for it

rasa_core_sdk/forms.py
erohmensing: yup

rasa_core_sdk/forms.py
erohmensing: I'll check again in a run to make sure, but it came from here:
```
    def _activate_if_required(self, tracker):
        # type: (Tracker) -> List[Dict]
        """Return `Form` event with the name of the form
            if the form was called for the first time"""

        if tracker.active_form.get("name") is not None:
            logger.debug("The form '{}' is active".format(tracker.active_form))
        else:
            logger.debug("There is no active form")

        if tracker.active_form.get("name") == self.name():
            return []
        else:
            logger.debug("Activated the form '{}'".format(self.name()))
            return [Form(self.name())]
```
and here:
```
class Form(Event):
    """If `name` is not None: activates a form with `name`
        else deactivates active form
    """
    type_name = "form"
```

rasa_core_sdk/forms.py
Ghostvv: can we reformat it? and put `and ...` on the new line?

rasa_core_sdk/forms.py
Ghostvv: but `Form` is not `ActionExecuted` so it cannot be `latest_action_name`, also it is not part of the tracker yet, because it is returned by the very same call to `FormAction`

rasa_core_sdk/forms.py
erohmensing: I tried 
```
elif (
      tracker.latest_action_name == "action_listen"
      and tracker.active_form.get("validate", True)
):
``` 
and no dice, black keeps changing it back

rasa_core_sdk/forms.py
erohmensing: if we call it from inside validate, it will re-validate all the slots it's already validated each time, no? I think it makes sense to keep this within the activation logic

rasa_core_sdk/forms.py
erohmensing: You're totally right. Sorry about that. Here's what I suggest instead:
[This function](https://github.com/RasaHQ/rasa_core_sdk/blob/master/rasa_core_sdk/forms.py#L418) will only return anything but `[]` if the form becomes activated, correct? Then I would do:
```python3
        # activate the form
        events = self._activate_if_required(tracker)
        # validate pre-filled slots
        if events != [] and tracker.active_form.get("validate", True):
            events.extend(self._validate_prefilled_slots(dispatcher, tracker, domain))
```
I've checked and this does actually go into the `validate_prefilled_slots()` as expected. My only concern is that that activation function looks quite weird, i would expect it to `return []` after the logger statements in the first logic chunk too. Does this make sense to you?

rasa_core_sdk/forms.py
Ghostvv: ^ above is not good for black @tabergma

rasa_core_sdk/forms.py
Ghostvv: that's should be fine, I think, the `else` is triggered when the form is activated. I don't really like checking for empty event list, because we (or actually anyone in subclass) might change the return statement

rasa_core_sdk/forms.py
tabergma: If black formats the line this way, we cannot do much about it. Sorry.

tests/test_forms.py
```diff
@@ -725,6 +725,58 @@ def validate_some_slot(self, value, dispatcher, tracker, domain):
...
+        Form("some_form"),
+        SlotSet("some_slot", "validated_value"),
+        SlotSet("some_other_slot", "some_other_value"),
+    ]
+
+    events.extend(
+        form._validate_if_required(dispatcher=None, tracker=tracker, domain=None)
+    )
+    # check that entities picked up in input overwrite prefilled slots
+    assert events == [
```
erohmensing: @Ghostvv here i added an extra check because i wanted to make sure that we had ideal behavior:
If `some_slot` is pre-filled, we validate the value of whatever is in the slot during activation. If we also picked up the entity `some_slot` from the input, that takes precedence and gets validated too. In this case, even though the value pre-filled slot was valid, the user provided a non-valid entity in the input, so the slot was set to none. I think this makes sense because if the user is trying to input information for that entity, then it should take whatever they input and not whatever was in the pre-filled slot. Does that make sense to you?

erohmensing: I suppose the only issue is that in order for it to overwrite, the timestamp of the event has to be later (pytest doesn't put timestamps on it). when do the timestamps actually get into the events -- will an event returned by the activate function correctly have a timestamp earlier than one returned by the validate function?

Ghostvv: I don't think timestamp is important here

erohmensing: I just mean that the timestamp would be the one to decide if `some_slot` ends up getting set to `validated_value` or `None`, right?

Ghostvv: sorry, I'm not sure I understand what you mean, if there are prefilled slots, it will validate them, then if there are entities, it will call validate again, then in `tracker.events` they will be executed based on the order in the list of events returned by form, and since second validate is called later it should override previous changes

tests/test_forms.py
```diff
@@ -725,6 +725,67 @@ def validate_some_slot(self, value, dispatcher, tracker, domain):
...
+        Form("some_form"),
+        SlotSet("some_slot", "validated_value"),
+        SlotSet("some_other_slot", "some_other_value"),
+        SlotSet("some_slot", None),
+    ] or events == [
```
Ghostvv: why `or`? I don't think there is any stochasticity

erohmensing: that's what i thought too, and then the travis test failed because some_other_slot came before some_slot

rasa_core_sdk/forms.py
```diff
@@ -425,8 +436,21 @@ def _list_intents(
...
+        )
+        logger.debug(
+            "No slots left to request, all required slots are filled:\n{}".format(
+                slot_values
+            )
+        )
+
+    def _activate_if_required(self, dispatcher, tracker, domain):
+        # type: (CollectingDispatcher, Tracker, Dict[Text, Any]) -> List[Dict]
         """Return `Form` event with the name of the form
```
Ghostvv: please add that prefilled slots will be validated in description

rasa_core_sdk/forms.py
```diff
@@ -439,7 +463,25 @@ def _activate_if_required(self, tracker):
...
+            prefilled_slots = {}
+            for slot_name in self.required_slots(tracker):
+                if not self._should_request_slot(tracker, slot_name):
+                    prefilled_slots[slot_name] = tracker.get_slot(slot_name)
+
+            if prefilled_slots:
+                logger.debug(
+                    "Validating pre-filled required slots: {}".format(prefilled_slots)
+                )
+                validation_events = self.validate_slots(
```
Ghostvv: let's just call it events like everywhere
```suggestion
                events.extend(self.validate_slots(
```

rasa_core_sdk/forms.py
```diff
@@ -439,7 +463,25 @@ def _activate_if_required(self, tracker):
             return []
         else:
             logger.debug("Activated the form '{}'".format(self.name()))
-            return [Form(self.name())]
+
```
Ghostvv: let's put here
```
events = [Form(self.name())]
```

rasa_core_sdk/forms.py
```diff
@@ -439,7 +463,25 @@ def _activate_if_required(self, tracker):
...
+            if prefilled_slots:
+                logger.debug(
+                    "Validating pre-filled required slots: {}".format(prefilled_slots)
+                )
+                validation_events = self.validate_slots(
+                    prefilled_slots, dispatcher, tracker, domain
+                )
+            else:
+                logger.debug("No pre-filled required slots to validate.")
+                validation_events = []
```
Ghostvv:
```suggestion
```

rasa_core_sdk/forms.py
```diff
@@ -439,7 +463,25 @@ def _activate_if_required(self, tracker):
...
+
+            return [Form(self.name())] + validation_events
```
Ghostvv:
```suggestion
            return events
```

erohmensing:
god bless. Sorry for the 5000 review requests, you're a champ

Ghostvv:
no worries, that's how it is usually

PR #81, opened 2019-05-23
re-add requests to requirements

erohmensing:


erohmensing:
maybe this branch should actually be the start of 1.0.x branch?

PR #86, opened 2019-06-27
Correct the latest restart index from events

psds01:
**Proposed changes**:
- The current `tracker` checks for a restart event by comparing against `restarted`, change it to ~~`action_restart`~~ `restart`

- [ ] updated the changelog

### DIFF
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -132,10 +132,11 @@ class Tracker(object):
 
         If the conversation has not been restarted, ``0`` is returned."""
 
-        for i, event in enumerate(reversed(self.events)):
-            if event.get("name") == "action_restart":
-                return len(self.events) - i
-        return 0
+        idx = 0
+        for i, event in enumerate(self.events):
+            if event.get("event") == "restarted":
+                idx = i + 1
+        return idx
 
     def events_after_latest_restart(self):
         # type: () -> List[dict]
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -133,7 +133,7 @@ class Tracker(object):
         If the conversation has not been restarted, ``0`` is returned."""
 
         for i, event in enumerate(reversed(self.events)):
-            if event.get("name") == "restart":
+            if event.get("event") == "restart":
                 return len(self.events) - i
         return 0
 
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -132,11 +132,10 @@ class Tracker(object):
 
         If the conversation has not been restarted, ``0`` is returned."""
 
-        idx = 0
-        for i, event in enumerate(self.events):
-            if event.get("event") == "restarted":
-                idx = i + 1
-        return idx
+        for i, event in enumerate(reversed(self.events)):
+            if event.get("name") == "action_restart":
+                return len(self.events) - i
+        return 0
 
     def events_after_latest_restart(self):
         # type: () -> List[dict]
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -134,7 +134,7 @@ class Tracker(object):
 
         idx = 0
         for i, event in enumerate(self.events):
-            if event.get("event") == "restarted":
+            if event.get("event") == "restart":
                 idx = i + 1
         return idx
 
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -133,7 +133,7 @@ class Tracker(object):
         If the conversation has not been restarted, ``0`` is returned."""
 
         for i, event in enumerate(reversed(self.events)):
-            if event.get("event") == "restart":
+            if event.get("name") == "restart":
                 return len(self.events) - i
         return 0
 
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -133,7 +133,7 @@ class Tracker(object):
         If the conversation has not been restarted, ``0`` is returned."""
 
         for i, event in enumerate(reversed(self.events)):
-            if event.get("name") == "action_restart":
+            if event.get("name") == "restart":
                 return len(self.events) - i
         return 0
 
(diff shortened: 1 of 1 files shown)


psds01:
@akelad @tmbo @Ghostvv

akelad:
it's not `action_restart` -- see here: https://github.com/RasaHQ/rasa-sdk/blob/master/rasa_sdk/events.py#L42

But @Ghostvv i guess it should indeed be `restart` and not `restarted` right?

psds01:
Great point @akelad!
You are 50% correct, I am 50% correct.
The reason I say `action_restart` is because of [this](https://github.com/RasaHQ/rasa/blob/3e6700d4147b5d60df6fd2399061c7e9daedb336/rasa/core/actions/action.py#L36).

I am correct if the `restart` event is inside the `rasa.core`, you are correct if the `restart` event is inside the `rasa_sdk`.

so I guess the solution(s) here is(are): 
1. make `rasa_sdk` *events* consistent with the `rasa.core` *events* (_not_ `class` types) means : no checking for `isinstance(event, EventType)` as can be found [here](https://github.com/RasaHQ/rasa/blob/3e6700d4147b5d60df6fd2399061c7e9daedb336/rasa/core/trackers.py#L244). Instead check with `event.type_name == "restarted"` inside `core`.
2. decouple the naming of events (as is the case currently with `restart` at least) and make `rasa_sdk` handle for multiple names(sdk)/types(core) of events: 
`event.get("event") == "restarted" or (event.get("name") =="action_restart")` inside `sdk`.
First condition to handle events generated by `sdk`, second to handle events generated by `core`.

Which one makes more sense? Am I missing something?
@akelad @Ghostvv

akelad:
the tracker in the sdk isn't an exact copy of the tracker in Core though. And the actions will only ever have access to the sdk tracker, therefore the `restart` event is correct

psds01:
`the tracker in the sdk isn't an exact copy of the tracker in Core though.` Interesting! I wonder why is that though?

akelad:
the sdk runs completely separately to Core and so it needs its own copy of it in order to be able to access it. It's different because it doesn't need all features of the Core one and so it's a bit more compact

psds01:
Well, let me give you little bit of context as to how I came across this issue and why sdk tracker and core tracker not being exact copy causes some (physical : writing more `if`s) trouble for me.

I am building a goal-oriented bot. In there, the bot comes across two ways a `restart` event can happen:

1. SDK: The bot has achieved the goal. Now it's time to "reset the clock" aka emit a `restart` event from the last action that I have in my pipeline, because the bot is not trained to handle post-goal-completion conversation.
2. CORE: The user manually restarts the conversation by typing `/restart` or by saying something like "please restart the conversation I am lost" or something to that effect. In this case, the NLU will identify the intent here as `restart` and the mapping policy will do its job of "triggering" `action_restart`.

This is how `restart` can happen and why I need them.

Now I have two utilities: 

1. First utility works on sdk tracker in `actions.py`. This checks for "some information" which is relevant only in the current conversation. Meaning it searches `sdk tracker` after **the latest restart event** (no matter how the tracker was "restarted" : by sdk actions or nlu intents)
2. Second utility works on core tracker in `rasa/core/processor.py`. This also checks for "some information" which is only relevant in the current conversation. Meaning it searches `core tracker` after **the latest restart event** (no matter how the tracker was "restarted" : sdk actions or nlu intents)

Now you can see, if I am searching for "XYZ" in `sdk tracker` and also in the `core tracker`, and **if** they are not in sync, I could get in real trouble. Now assume that the "XYZ" is OTP sent to the user for some purpose.

It's a big surprise to me that they are not exactly same, albeit sdk being a bare version of core tracker. I was relying on that. Maybe I missed the disclosure that they are not exactly the same?

Do you have any ideas about solving this (little/huge) problem? I mean sure, I can type in few conditions to handle two "different" trackers, but intuitively that does not seem like a good design.

akelad:
What are you trying to achieve with those two things? I'm not sure I understand

psds01:
I am putting some constraints on NLU. I want the NLU to give me an `intent` which is a correct response to the last question the bot asked or to give me `out_of_scope`. This increases precision of my "new" NLU to almost `1`.

To find out the intent type that I want: I need access to the tracker after the latest `restart` (by `core` or by `sdk`)

akelad:
I'm not sure I understand the problem with using the two separate trackers though. Also, could you please revert all your changes apart from the rename to `restart`? the original version seemed more readable to me

psds01:
Changes reverted and correct name of the event used.

Looks like there's some more convincing that needs to take place before on-boarding anybody with this idea(keeping both the tracker in sync in terms of design flexibility of the bot and custom actions) or [this](https://github.com/RasaHQ/rasa/pull/3681/) or [this](https://github.com/RasaHQ/rasa/issues/3015).
1. The performance of NLU on its own should not be expected to be too high. And in my case, it's not that great. To overcome this, I **constrain** the NLU parser by providing one extra parameter which tells it what I am looking for, else the parser provides `out_of_scope`. This way I get almost perfect results. 

Instead of providing just one parameter, in future, _and in general_, I could provide the complete tracker(since latest restart) as "context" which is discussed in [this](https://github.com/RasaHQ/rasa/pull/3681/) and [this](https://github.com/RasaHQ/rasa/issues/3015).

2. To make conversation with a user a little bit *human-like*, I need custom actions to utter helpful messages to make users understand what is needed. If the sdk event names are exactly the same as in core, I could write a module/third-party package which will be used by both sdk and core to give different types of "context features".

Anyway, for now I will write two `if` conditions to solve my problem.

Thanks for your time @akelad  :)

akelad:
@Ghostvv just to double check, there's no reason the event would be called `restarted` rather than `restart`?

Ghostvv:
it is `restart` in both [`rasa`](https://github.com/RasaHQ/rasa/blob/84afa6e61c8d109bf6fb515268b937f5315c8f9b/rasa/core/events/__init__.py#L472) and [`rasa-sdk`](https://github.com/RasaHQ/rasa-sdk/blob/b18f088446e405608b74193e84c0d82ddfe6af47/rasa_sdk/events.py#L43)

PR #88, opened 2019-07-12
add utter_image_url

rajithacharith:
**Proposed changes**:
- When trying to send image from custom actions there is no function in the dispatcher to send image as a URL. But in rasa core, it supports the image sending and only needs to send image url formated as {"image" : image_url }. Added utter_image_url function and implementation to CollectingDispatcher class. 

- [+] made PR ready for code review
- [- ] added some tests for the functionality
- [- ] updated the documentation
- [- ] updated the changelog

### DIFF
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -116,6 +116,14 @@ class CollectingDispatcher(object):
 
         self.messages.append(json_message)
 
+    def utter_image_url(self, image, **kwargs):
+        """ sends image url to the output channel as image"""
+
+        json_message = {"image": image}
+        json_message.update(kwargs)
+
+        self.messages.append(json_message)
+
 
 class ActionExecutor(object):
     def __init__(self):
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -116,11 +116,8 @@ class CollectingDispatcher(object):
 
         self.messages.append(json_message)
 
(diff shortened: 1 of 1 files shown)


akelad:

rasa_sdk/executor.py
```diff
@@ -116,6 +116,14 @@ def utter_custom_json(self, json_message, **kwargs):
 
         self.messages.append(json_message)
 
+    def utter_image_url(self, image, **kwargs):
```
erohmensing: Can you add types (as above) please?

rasa_sdk/executor.py
```diff
@@ -116,6 +116,14 @@ def utter_custom_json(self, json_message, **kwargs):
...
+        """ sends image url to the output channel as image"""
```
erohmensing:
```suggestion
        """Sends url of image attachment to the output channel."""
```

rasa_sdk/executor.py
```diff
@@ -116,6 +116,14 @@ def utter_custom_json(self, json_message, **kwargs):
...
+
+        json_message = {"image": image}
```
erohmensing: please change these `json_message` to `message` (`utter_custom_json` is a little special -- the rest are just `message`s.)

rajithacharith:
Thanks for the quick reply :)
I updated the code as requested

PR #96, opened 2019-08-08
Add action to query knowledge base

tabergma:
related PR https://github.com/RasaHQ/rasa/pull/4210

**Proposed changes**:
- Add action to query a knowledge base. The action can retrieve objects of a specific type and can answer questions about attributes of those objects. It is also able to resolve mentions like "first . one" or "last". Other mentions like "it" or "that restaurant" would be resolved to the last detected object per default.
- Add interface for `KnowledgeBase` and the implementation for an `InMemoryKnowledgeBase` that users can use to build up a simple knowledge base using a json.
- Internal slots, such as `knowledge_base_listed_objects`, are set automatically by rasa as soon as the default action name is listed in the domain file (prefix `knowledge_base_`)

What does the user need to do:
1. Create a json file that contains the data. For example:
```
{
    "restaurant": [
        {"id": 0, "name": "PastaBar", "cuisine": "Italian", "wifi": "False"},
        {"id": 1, "name": "Berlin Burrito Company", "cuisine": "Mexican", "wifi": "True"},
        {"id": 2, "name": "I due forni", "cuisine": "Italian", "wifi": "False"}
    ],
    "hotel": [
        {
            "id": 0,
            "name": "Hilton",
            "price-range": "expensive",
            "breakfast-included": "True",
            "city": "Berlin"
        },
        {
            "id": 1,
            "name": "Hilton",
            "price-range": "expensive",
            "breakfast-included": "True",
            "city": "Frankfurt am Main"
        },
        {
            "id": 2,
            "name": "B&B",
            "price-range": "mid-range",
            "breakfast-included": "True",
            "city": "Berlin"
        },
        {
            "id": 3,
            "name": "Berlin Wall Hostel",
            "price-range": "cheap",
            "breakfast-included": "False",
            "city": "Berlin"
        },
        {
            "id": 4,
            "name": "City Hotel",
            "price-range": "expensive",
            "breakfast-included": "False",
            "city": "Berlin"
        },
        {
            "id": 5,
            "name": "Jugendherberge",
            "price-range": "cheap",
            "breakfast-included": "True",
            "city": "Berlin"
        },
        {
            "id": 6,
            "name": "Berlin Hotel",
            "price-range": "mid-range",
            "breakfast-included": "False",
            "city": "Berlin"
        }
    ]
}
```
2. Add a custom actions file. It has to initialize a knowledge base and pass it along to the knowledge base action. For example:
```
class ActionMyKB(ActionQueryKnowledgeBase):
    def __init__(self):
        knowledge_base = InMemoryKnowledgeBase.load("data.json")

        knowledge_base.set_representation_function_of_object(
            "hotel", lambda obj: obj["name"] + " (" + obj["city"] + ")"
        )

        super().__init__(knowledge_base)
```
2. Add an intent with examples to the NLU data. The knowledge base action makes use of the following entities `object_type`, `mention`, and `attribute`. The user has to mark those in the training data. For example:
```
## intent:query_knowledge_base
- what [restaurants](object_type:restaurant) can you recommend?
- list some [restaurants](object_type:restaurant)
- can you show me some [restaurant](object_type:restaurant) options
- does the [first](mention:1) one has [wifi](attribute)?
- what [cuisine](attribute) is it?
- does the [last](mention:LAST) one offer [breakfast](attribute:breakfast-included)?
- do you know the [cuisine](attribute) of [that one](mention)?
- do you have any [mexican](cuisine) [restaurants](object_type:restaurant)?
- can you name some [restaurants](object_type:restaurant) please?
- do you know what [cuisine](attribute) the [last one](mention:LAST) has?
- does [PastaBar](restaurant) have [wifi](attribute)?
- what is the [cuisine](attribute) of [Berlin Burrito Company](restaurant)?
- what is with [I due forni](restaurant)?
- name some [hotels](object_type:hotel) in [Berlin](city)
- what are [hotels](object_type:hotel) in [Berlin](city)
- does the [second one](mention:2) [include breakfast](breakfast-included)?
- what is the [price range](price-range) of the [second](mention:2) hotel?
```
3. Add at least one story that makes use of the newly defined intent and the knowledge base action, called `action_query_knowledge_base` per default
4. Add all needed slots, intents, actions to the domain file

- [ ] updated the documentation (updated in https://github.com/RasaHQ/rasa/pull/4210)
- [x] updated the changelog


tabergma:
@amn41 Do you have any early feedback on this? I tired to minimize the work of the user. I was trying to use some configuration (e.g. in config.py) to actually define the action + knowledge base. However, that is not that easily done and would involve some hacks. But as it is still kind of experiential, I left it as it is. Let me know if you have an idea/comment. (cc @tmbo)

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,283 @@
...
+class ActionKnowledgeBase(Action):
+    """
+    Abstract knowledge base action that can be inherited to create custom actions
+    that are able to interact with the knowledge base.
+    """
+
+    def __init__(self, knowledge_base):
+        self.knowledge_base = knowledge_base
+
+    def name(self):
```
wochinge: you can skip these two definitions since the parent class should handle this

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,283 @@
...
+    def run(self, dispatcher, tracker, domain):
+        raise NotImplementedError("An action must implement its run method")
+
+    def _get_entity(self, tracker):
+        """
+        Get the name of the entity the user referred to. Either the NER detected the
+        entity and stored its name in the corresponding slot or the user referred to
+        the entity by any kind of mention, such as "first one" or "it".
+
+        :param tracker: Tracker
```
wochinge: Can we use the google style for the doc strings?

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,283 @@
...
+                "Sorry, I did not get that. Can you please rephrase?"
+            )
+            return []
+
+        if not attribute or new_request:
+            return self._query_entities(dispatcher, tracker)
+        elif attribute:
+            return self._query_attribute(dispatcher, tracker)
+
+        dispatcher.utter_message("Sorry, I did not get that. Can you please rephrase?")
```
wochinge: think we should utter a template here

rasa_sdk/knowledge_base/storage.py
```diff
@@ -0,0 +1,136 @@
...
+    SCHEMA_KEYS_KEY,
+    SCHEMA_KEYS_ATTRIBUTES,
+    SCHEMA_KEYS_REPRESENTATION,
+]
+
+
+class KnowledgeBase(object):
+    def __init__(self, schema):
+        self.ordinal_mention_mapping = {
+            "one": lambda l: l[0],
```
wochinge: isn't there a library for that?

wochinge: also duckling should be able to extract that and pass it to us, right?

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,283 @@
...
+
+        representation_function = self.knowledge_base.schema[entity_type][
+            SCHEMA_KEYS_REPRESENTATION
+        ]
+
+        dispatcher.utter_message(
+            "Found the following entities of type '{}':".format(entity_type)
+        )
+        for i, e in enumerate(entities):
+            dispatcher.utter_message("{}: {}".format(i + 1, representation_function(e)))
```
wochinge: I think we should use templates for all these `dispatches`

wochinge: enumerate can also start with 1 🙂

rasa_sdk/knowledge_base/storage.py
amn41: yes, but adding duckling service as a dependency for such a simple thing would be overkill IMO

amn41: as @Ghostvv  suggested, could we just use synonyms for this mapping? so `the [first](mention:1) one?` then the `ordinal_mention_mapping` dict can be made language agnostic.

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+        return tracker.get_slot(SLOT_LAST_ENTITY)
+
+    def _resolve_mention(self, tracker):
+        """
+        Resolves a mention of an entity, such as first, to the actual entity.
+        If multiple entities are listed during the conversation, the entities
+        are stored in the slot 'knowledge_base_entities' as a list. We resolve the
+        mention, such as first, to the list index and retrieve the actual entity.
+        If the mention is not an ordinal mention, but some other reference, we just
+        assume the user is referring to the last mentioned entity in the conversation.
```
amn41: following on from our discussion, I feel we are potentially causing a lot of confusion by referring to the objects in the knowledge base as entities. Maybe we should call them something like 'objects' or 'items' or 'nodes' or 'mentions'

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+        Returns: a string that represents the entity
+        """
+        representation_func = self.knowledge_base.schema[entity_type][
+            SCHEMA_KEYS_REPRESENTATION
+        ]
+        return representation_func(entity)
+
+    def _get_attributes_of_entity(self, entity_type, tracker):
+        """
+        Checks if the NER found any value for all attributes of the given entity type.
```
amn41: this comment is hard to parse. can we make it more concrete with an example ?

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+
+class ActionQueryKnowledgeBase(ActionKnowledgeBase):
+    """
+    Action that queries the knowledge base for entities and attributes of an entity.
+    The action needs to be inherited and the knowledge base needs to be set.
+    In order to actually query the knowledge base you need to:
+    - create your knowledge base
+    - add mandatory slots to the domain file: 'entity_type', 'attribute', 'mention'
+    - create an intent that triggers this action
+    - mark all needed entities in the NLU data, such as 'entity_type'
```
amn41: - create NLU data where the required entities are annotated. For example, ...

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+    - create a story that includes this action
+    - add created intent and action to domain file
```
amn41:
```suggestion
    - add the intent and action to domain file
```

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+    def run(self, dispatcher, tracker, domain):
+        entity_type = tracker.get_slot(SLOT_ENTITY_TYPE)
+        last_entity_type = tracker.get_slot(SLOT_LAST_ENTITY_TYPE)
+        attribute = tracker.get_slot(SLOT_ATTRIBUTE)
+
+        new_request = entity_type != last_entity_type
+
+        if not entity_type:
+            dispatcher.utter_template(
+                "Sorry, I did not get that. Can you please rephrase?", tracker
```
amn41: maybe implement this as a function that people can overwrite easily

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+            )
+
+            slots = [SlotSet(SLOT_MENTION, None)]
+            return slots + self._reset_attribute_slots(entity_type, tracker)
+
+        value = self.knowledge_base.get_attribute_of(entity_type, entity, attribute)
+
+        # utter response
+        if value:
```
amn41: maybe also wrap this in a function that's overridable so people can change the text without copying the logic in this function

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+            return attributes
+
+        for attr in self.knowledge_base.schema[entity_type][SCHEMA_KEYS_ATTRIBUTES]:
+            attr_val = tracker.get_slot(attr)
+            if attr_val is not None:
+                attributes.append({"name": attr, "value": attr_val})
+
+        return attributes
+
+    def _reset_attribute_slots(self, entity_type, tracker):
```
amn41: it's not obvious to me why we need this

rasa_sdk/knowledge_base/storage.py
tabergma: Yes, good idea. Did that already with the `entity_type` and `attributes`.

rasa_sdk/knowledge_base/actions.py
tabergma: If the user is saying something like "show me all restaurants with Italian cuisine", the NER should detect `restaurant` as `entity_type` and `Italian` as `cuisine` entity. So, we should filter the restaurant entities by their cuisine (= Italian). 
When listing entities, we check what attributes are detected by the NER. We take all attributes that are set, e.g. `cuisine=Italian`. If we don't reset the attribute slots after the request is done and the next utterance of the user would be, for example, "list all restaurants that have wifi", we now have two attribute slots set: `wifi` and `cuisine`. Thus, we would filter all restaurants for two attributes now: `wifi=True` and `cuisine=Italian`. To avoid that we reset the attribute slots once the request is done.

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,290 @@
...
+        # entity mentioned in the conversation
+        return tracker.get_slot(SLOT_LAST_ENTITY)
+
+    def _resolve_mention(self, tracker):
+        """
+        Resolves a mention of an entity, such as first, to the actual entity.
+        If multiple entities are listed during the conversation, the entities
+        are stored in the slot 'knowledge_base_entities' as a list. We resolve the
+        mention, such as first, to the list index and retrieve the actual entity.
+        If the mention is not an ordinal mention, but some other reference, we just
```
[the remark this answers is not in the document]
tabergma: `it` or `that restaurant` (added it to the doc string)

rasa_sdk/knowledge_base/actions.py
amn41: ok! I suspected this was the case, but that should be documented in the docstring & docs

rasa_sdk/knowledge_base/storage.py
```diff
@@ -18,25 +18,18 @@
...
-            "four": lambda l: l[3],
-            "fourth": lambda l: l[3],
             "4": lambda l: l[3],
+            "5": lambda l: l[4],
+            "6": lambda l: l[5],
+            "7": lambda l: l[6],
+            "8": lambda l: l[7],
+            "9": lambda l: l[8],
+            "10": lambda l: l[9],
             "any": lambda l: random.choice(list),
```
amn41: maybe for any and last we should replace these with `ANY`, `ALL`, `LAST` to indicate to users that these are symbols ?

rasa_sdk/knowledge_base/actions.py
tabergma: changed `entities` to `objects`

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,372 @@
...
+
+
+class ActionKnowledgeBase(Action):
+    """
+    Abstract knowledge base action that can be inherited to create custom actions
+    that are able to interact with a knowledge base.
+    """
+
+    def __init__(self, knowledge_base):
+        self.knowledge_base = knowledge_base
```
Ghostvv: if I remember correctly, action is init every time, shouldn't this knowledge_base be a part of the tracker?

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        representation_func = self.knowledge_base.schema[object_type][
+            SCHEMA_KEYS_REPRESENTATION
+        ]
+        return representation_func(object_dict)
+
+    def _get_attributes_of_object(self, tracker, object_type):
+        """
+        If the user mentioned one or multiple attributes of the provided object_type in
+        his utterance, we extract all attribute values from the tracker and put them
```
Ghostvv:
```suggestion
        an utterance, we extract all attribute values from the tracker and put them
```

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,372 @@
...
+        cuisine.", the NER should detect "restaurant" as "object_type" and "Italian" as
+        "cuisine" object. So, we should filter the restaurant objects in the
+        knowledge base by their cuisine (= Italian). When listing objects, we check
+        what attributes are detected by the NER. We take all attributes that are set,
+        e.g. cuisine = Italian. If we don't reset the attribute slots after the request
+        is done and the next utterance of the user would be, for example, "List all
+        restaurants that have wifi.", we would have two attribute slots set: "wifi" and
+        "cuisine". Thus, we would filter all restaurants for two attributes now:
+        wifi = True and cuisine = Italian. However, the user did not specify any
+        cuisine in his request. To avoid that we reset the attribute slots once the
```
Ghostvv:
```suggestion
        cuisine in the request. To avoid that we reset the attribute slots once the
```

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def utter_rephrase(self, dispatcher):
+        """
+        Utters a response to the user that indicates that something went wrong. It
+        asks the user to rephrase his request.
+
+        Args:
+            dispatcher: the dispatcher
+        """
+        dispatcher.utter_message("Sorry, I did not get that. Can you please rephrase?")
```
Ghostvv: Let's not hardcode the messages, but rather ask a user to create utter templates with certain names

rasa_sdk/knowledge_base/actions.py
```diff
@@ -0,0 +1,372 @@
...
+
+SLOT_MENTION = "mention"
+SLOT_OBJECT_TYPE = "object_type"
+SLOT_ATTRIBUTE = "attribute"
+SLOT_LISTED_OBJECTS = "knowledge_base_listed_objects"
+SLOT_LAST_OBJECT = "knowledge_base_last_object"
+SLOT_LAST_OBJECT_TYPE = "knowledge_base_last_object_type"
+
+
+class ActionKnowledgeBase(Action):
```
Ghostvv: it seems to me that it is not really an `Action` but rather a collection of helper methods, because it doesn't have `run` method. I'd rather merge this with another action. Even though it'll be very large, I think it'll be less confusing for a user

tabergma: I kept it separate so that users could inherit this Action and implement more use cases. That way they could use our default implementation along their one implementations.

rasa_sdk/knowledge_base/actions.py
tabergma: Not sure about when an action is init. But adding knowledge_base to the tracker would complicate things quite a bit. Maybe @tmbo or @wochinge knows.

rasa_sdk/knowledge_base/actions.py
Ghostvv: but cannot they inherit the other one and implement their implementation?

rasa_sdk/knowledge_base/actions.py
Ghostvv: I remember when we did `FormAction` one of the problems was that it should be stateless

rasa_sdk/knowledge_base/actions.py
tabergma: yeah... but I actually don't like that. I guess it is more clear if you keep it separate.

Ghostvv: I see, ok. Can we add some run method there? I think it's a bit not clear how to use it on its own

tabergma: I removed it as @wochinge pointed out that they are not needed. But can add them again.

Ghostvv: I don't know, but if this action can be used by its own, I'd put `run` action there

rasa_sdk/knowledge_base/actions.py
tabergma: I discussed it with @tmbo. Should not be a problem. The action is initialized just once.

rasa_sdk/knowledge_base/actions.py
tabergma: Removed the abstract action and added all helper functions to a `utils` package.

rasa_sdk/knowledge_base/actions.py
Ghostvv: wait, so custom action can contain instance variables?

Ghostvv: We discussed it with tmbo, now I see why it's fine and instance variables are not

tabergma:
I have one open question: The advantage of knowledge bases is that you can exchange/update the underlying data and the bot automatically picks those up. For the In-Memory KB we load the data just once at the beginning. How can we overcome this? Should we load the data on every request?

PR #108, opened 2019-10-30
Remove duplicate slot validation

yaput:
**Proposed changes**:
- Removed slot validation inside `_activate_if_required()`

- [ ] updated the changelog

### DIFF
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -467,22 +467,6 @@ class FormAction(Action):
             logger.debug("Activated the form '{}'".format(self.name()))
             events = [Form(self.name())]
 
-            # collect values of required slots filled before activation
-            prefilled_slots = {}
-            for slot_name in self.required_slots(tracker):
-                if not self._should_request_slot(tracker, slot_name):
-                    prefilled_slots[slot_name] = tracker.get_slot(slot_name)
-
-            if prefilled_slots:
-                logger.debug(
-                    "Validating pre-filled required slots: {}".format(prefilled_slots)
-                )
-                events.extend(
-                    self.validate_slots(prefilled_slots, dispatcher, tracker, domain)
-                )
-            else:
-                logger.debug("No pre-filled required slots to validate.")
-
             return events
 
     def _validate_if_required(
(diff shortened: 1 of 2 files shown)


yaput:
Hi, i'm removing validation code inside the `_activate_if_required()` inside forms.py why? Because on forms.py inside `run()` line 519, we will validate the slot again. 
I tried to test before, and put `dispatcher.utter_message()` inside `validate_{slot_name}()` when it's not satisfied the validation and it will be executed twice. Is it expected?

akelad:
hi @yaput this shouldn't be removed, because it is there to validate any slots that may have been filled prior to the activation of the form. E.g. if the user says "I want a chinese restaurant" and this activates the form, that cuisine entity needs to be validated.

yaput:
hi @akelad yes, i agree with that, but it will be validated anyway right? After _activate_if_required, then the next code is _validate_if_required. Because right now it will execute the validate_{slot_name} twice, and my dispatcher.utter_message() to notify the user will run twice. I tried to use that case, and it will validate the slot anyway. I tried it on my local machine

akelad:
hm actually i'm not too sure now, @Ghostvv any reason why we double validate there?

Ghostvv:
`_activate_if_required` validates profiled slots. `_validate_if_required` validates slots that are set inside the last user input. But it validates different things: first time the slot, the second time the entity.  If a user provides the slot that is already filled or you fill the slots automatically with entities with the same name, it will be called it twice.

yaput:
Hi @Ghostvv thanks for the explanation, if they are validating different thing, why they use the same function `validate()`?

yaput:
Hi @Ghostvv thanks a lot, do you have any suggestion how to avoid invoking the `validate_{slot_name}` twice? Right now i'm using the same name for slot and entity, and i set the slot using entity `/intent_name{"entity_1": "value1"}` then after that i activate the form, that's why it's executing twice. value1 based on user input.

Ghostvv:
I'd suggest to turn off auto filling of slots that are part of the form, as we do in formbot example: https://github.com/RasaHQ/rasa/blob/7fcfdd9eaa8704dc44ecbdb2927eb1a07b10d333/examples/formbot/domain.yml#L24

PR #112, opened 2019-11-05
Dispatcher methods

imLew:
**Proposed changes**:
- Replace the `CollectingDispatcher` methods with a single method. The former structure was based on the sdk's interactin with rasa_core and is no longer necessary.

Related to https://github.com/RasaHQ/rasa/pull/4687
(Guess I should also update the changelog there)

- [ ] updated the changelog

### DIFF
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -64,7 +64,7 @@ class CollectingDispatcher:
             "Use of `utter_elements` is deprecated. Use `utter_message` instead.",
             DeprecationWarning,
         )
-        self.utter_message(*elements)
+        self.utter_message(*elements, **kwargs)
 
     def utter_button_message(
         self, text: Text, buttons: List[Dict[Text, Any]], **kwargs: Any
@@ -76,7 +76,7 @@ class CollectingDispatcher:
             DeprecationWarning,
         )
 
-        self.utter_message(text=text, buttons=buttons)
+        self.utter_message(text=text, buttons=buttons, **kwargs)
 
     def utter_attachment(self, attachment: Text, **kwargs: Any) -> None:
         """Send a message to the client with attachments."""
@@ -85,7 +85,7 @@ class CollectingDispatcher:
             DeprecationWarning,
         )
 
-        self.utter_message(attachment=attachment)
+        self.utter_message(attachment=attachment, **kwargs)
 
     # noinspection PyUnusedLocal
     def utter_button_template(
@@ -103,7 +103,7 @@ class CollectingDispatcher:
             DeprecationWarning,
         )
 
-        self.utter_message(template=template, buttons=buttons)
+        self.utter_message(template=template, buttons=buttons, **kwargs)
 
     # noinspection PyUnusedLocal
     def utter_template(
@@ -115,7 +115,7 @@ class CollectingDispatcher:
             DeprecationWarning,
         )
 
-        self.utter_message(template=template)
+        self.utter_message(template=template, **kwargs)
 
     def utter_custom_json(self, json_message: Dict[Text, Any], **kwargs: Any) -> None:
         """Sends custom json to the output channel."""
@@ -124,7 +124,7 @@ class CollectingDispatcher:
             DeprecationWarning,
         )
 
-        self.utter_message(json_message=json_message)
+        self.utter_message(json_message=json_message, **kwargs)
 
     def utter_image_url(self, image: Text, **kwargs: Any) -> None:
         """Sends url of image attachment to the output channel."""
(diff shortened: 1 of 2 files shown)


erohmensing:
How did I never think to do this -- I love this 😃

rasa_sdk/executor.py
```diff
@@ -23,49 +23,69 @@ def __init__(self) -> None:
...
-
-    def utter_message(self, text: Text, **kwargs: Any) -> None:
-        """"Send a text to the output channel"""
-
-        message = {"text": text}
-        message.update(kwargs)
-
-        self.messages.append(message)
+        warnings.warn(
+            "Use of `utter_elements` is deprecated. " "Use `utter_message` instead.",
```
tabergma:
```suggestion
            "Use of `utter_elements` is deprecated. Use `utter_message` instead.",
```

rasa_sdk/executor.py
```diff
@@ -23,49 +23,69 @@ def __init__(self) -> None:
...
+            "Use of `utter_button_message` is deprecated. "
+            "Use `utter_message` instead.",
+            DeprecationWarning,
+        )
 
-        message = {"text": text, "buttons": buttons}
-        message.update(kwargs)
-
-        self.messages.append(message)
+        self.utter_message(text=text, buttons=buttons)
```
tabergma: What about `kwargs`? Same for the other methods.

rasa_sdk/executor.py
```diff
@@ -23,49 +23,69 @@ def __init__(self) -> None:
...
 
     def utter_attachment(self, attachment: Text, **kwargs: Any) -> None:
         """Send a message to the client with attachments."""
+        warnings.warn(
+            "Use of `utter_attachment` is deprecated. " "Use `utter_message` instead.",
```
tabergma:
```suggestion
            "Use of `utter_attachment` is deprecated. Use `utter_message` instead.",
```

rasa_sdk/executor.py
```diff
@@ -77,38 +97,43 @@ def utter_button_template(
...
-        self.messages.append(message)
+        self.utter_message(template=template, buttons=buttons)
 
     # noinspection PyUnusedLocal
     def utter_template(
         self, template: Text, tracker: Tracker, silent_fail: bool = False, **kwargs: Any
     ) -> None:
         """"Send a message to the client based on a template."""
+        warnings.warn(
+            "Use of `utter_template` is deprecated. " "Use `utter_message` instead.",
```
tabergma:
```suggestion
            "Use of `utter_template` is deprecated. Use `utter_message` instead.",
```

rasa_sdk/executor.py
```diff
@@ -77,38 +97,43 @@ def utter_button_template(
...
-        message = {"template": template}
-        message.update(kwargs)
-
-        self.messages.append(message)
+        self.utter_message(template=template)
 
     def utter_custom_json(self, json_message: Dict[Text, Any], **kwargs: Any) -> None:
         """Sends custom json to the output channel."""
+        warnings.warn(
+            "Use of `utter_custom_json` is deprecated. " "Use `utter_message` instead.",
```
tabergma:
```suggestion
            "Use of `utter_custom_json` is deprecated. Use `utter_message` instead.",
```

rasa_sdk/executor.py
```diff
@@ -77,38 +97,43 @@ def utter_button_template(
...
-        json_message = {"custom": json_message}
-        json_message.update(kwargs)
-
-        self.messages.append(json_message)
+        self.utter_message(json_message=json_message)
 
     def utter_image_url(self, image: Text, **kwargs: Any) -> None:
         """Sends url of image attachment to the output channel."""
+        warnings.warn(
+            "Use of `utter_image_url` is deprecated. " "Use `utter_message` instead.",
```
tabergma:
```suggestion
            "Use of `utter_image_url` is deprecated. Use `utter_message` instead.",
```

imLew: Thanks for these. Sorry you had to do this 😓, I will start checking strings after black runs :/

rasa_sdk/executor.py
imLew: Thanks for catching that 👍 I will pass on the kwargs
(Out of curiosity: do you know if they are ever used?)

tabergma: I have no idea 😄

imLew: better safe than sorry 😂 
just made the change

imLew:
> How did I never think to do this -- I love this 😃

When you're writing docs, but you're too lazy to write docs so you end up refactoring instead :P

erohmensing:
> When you're writing docs, but you're too lazy to write docs so you end up refactoring instead :P

I'm gonna put you in charge of more docs! 😉

PR #115, opened 2019-11-14
prevent non-vital commands from running as root

### LINKED ISSUE #114: __pycache__ is created as a root user, dockerfile should not run as priviledged user
**Rasa version**: 1.4.5 (not relevant for this issue)

**Rasa SDK version**: 1.4.0

**Python version**: 3.6

**Operating system** (windows, osx, ...): Docker

**Issue**:
When running the rasa-sdk container and loading the actions module via `entrypoint.sh`, `actions/__pycache__` is created as a root user.  Since this is typically a directory, attempts to delete `__pycache__` as a non-root user will fail with error below.  

Rasa's containers (in general) should not run as root.

```
rm: cannot remove 'actions/__pycache__/mappedActions.cpython-36.pyc': Permission denied
rm: cannot remove 'actions/__pycache__/slotChecks.cpython-36.pyc': Permission denied
rm: cannot remove 'actions/__pycache__/fillSlots.cpython-36.pyc': Permission denied
```
> Note: obviously, running `sudo rm -rf actions/__pycache` is able to clear this folder out.

```
docker-compose run action-server

rm -rf actions/__pycache__
```

**docker-compose.yml contents (abriged):**
```yml
version: 3.0
services:
  # ...
  action-server:
    image: rasa/rasa-sdk:1.4.0
    networks: ['rasa-network']
    ports:
      - '5055:5055'
    volumes:
      - ./actions:/app/actions
    command:
      - start
      - --actions
      - actions
```

ortsaCniveK:
**Proposed changes**:
- prevent `rasa-sdk` containers from running as root, using [this pattern](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b)
> fixes #114 

> Note: image file increased by only 1 MB after proposed change:
> ```sh
> REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
> test                1.0                 f97cbb6f4eb7        21 minutes ago      423MB
> rasa/rasa-sdk       1.4.0               aac88e91443f        3 weeks ago         422MB
> ```

- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,7 +25,7 @@ Changed
 
 Fixed
 -----
-- ``rasa-sdk`` containers will now not run as root
+- ``rasa-sdk`` containers will now not run as ``root``
 
 Removed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Changed
 
 Fixed
 -----
+- ``rasa-sdk`` containers will now not run as root
 
 Removed
 -------
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ COPY . .
 
 RUN pip install -e . --no-cache-dir
 
+RUN groupadd -g 1000 nonroot && \
+  useradd -r -u 1000 -g nonroot nonroot
+USER nonroot
+
 VOLUME ["/app/actions"]
 
 EXPOSE 5055
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY . .
 RUN pip install -e . --no-cache-dir
 
 RUN groupadd -g 1000 nonroot && \
-  useradd -r -u 1000 -g nonroot nonroot
+  useradd -r -u 1001 -g nonroot nonroot
 USER nonroot
 
 VOLUME ["/app/actions"]


akelad:

CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Changed
 
 Fixed
 -----
+- ``rasa-sdk`` containers will now not run as root
```
wochinge:
```suggestion
- ``rasa-sdk`` containers will now not run as ``root``
```

Dockerfile
```diff
@@ -20,6 +20,10 @@ COPY . .
 
 RUN pip install -e . --no-cache-dir
 
+RUN groupadd -g 1000 nonroot && \
+  useradd -r -u 1000 -g nonroot nonroot
```
wochinge: in our other images we are using `1001`, can we keep that consistent?
```suggestion
  useradd -r -u 1001 -g nonroot nonroot
```

Dockerfile
```diff
@@ -20,6 +20,10 @@ COPY . .
 
 RUN pip install -e . --no-cache-dir
 
+RUN groupadd -g 1000 nonroot && \
```
wochinge: Some container services (such as OpenShift) run with random users, but assume that everything can be accessed by the `root` group which in contrast to the `root` user does not have any special permissions (see https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html#openshift-enterprise-specific-guidelines -> Support Arbitrary User IDs). Hence, we should add the user to the root group and not create a new group.

wochinge: @ortsaCniveK have you seen this comment?

ortsaCniveK: Sorry, was caught up with the holidays and work.  The root reason was to try and clean up the permissions of the`__pycache__` folder that's created automatically as it has caused issues in mounting volumes as in [this issue](https://github.com/RasaHQ/rasa/issues/4653) , but I guess you can either suppress its creation [here by changing the line to `python -B -m rasa_sdk`](https://github.com/RasaHQ/rasa-sdk/blob/17eb70187555c8656336f9f21dcb2eb18655f455/entrypoint.sh#L14) (not necessarily ideal for some users due to performance implications) or just remember to `sudo rm -rf __pycache__` if someone runs into issues, which I don't think is intuitive either.

wochinge:
@ortsaCniveK Why did you close it? Would be great to get this in 🎉

Dockerfile
wochinge: you mean running `sudo rm -rf __pycache__` before they mount the volume?

ortsaCniveK: correct; I think that it could potentially be exclusively a Linux issue, as I've also had similar issues running interactive mode with failure to create `story_graph.dot` and the resulting interactive story files.

Perhaps we simply need to is just use the non-root `$UID` for the current user?  I'll investigate to see what we could change to do that.

wochinge: > Perhaps we simply need to is just use the non-root $UID for the current user?

Yep, let's try that. That's what we are currently also doing in the rasa image. Can you share how the permissions of your mounted directory look like?

wochinge:
@ortsaCniveK what's your state? Otherwise I'd close it for now.

ortsaCniveK:
I've changed the UID to match the other images, though I think this might need to be tabled for now.  I can revisit this later as I think there are cross-platform considerations that need to be kept in mind (e.g. I have to chmod the `models` folder if I run a training session from a docker image, but only on Linux).

I'm fine with closing for now :)

wochinge:
Ok, I'll close it for now and create an issue for it with a link to this PR so that we or some community members can wrap it up 👍

PR #127, opened 2019-11-28
use utter_message as recommended

erohmensing:
**Proposed changes**:
- Use utter_message as recommended. Will also update the example bots in rasa to that effect

- [ ] updated the changelog


imLew:
Is this PR done or will you include the changes to examples here as well @erohmensing?
Also since `text` is the first argument this does not change behaviour but only makes it more explicit right?

erohmensing:
Haha I can't update the rasa bots in this repo. Yes this would only make it more explicit. I guess the question is when you changed the functionality, did you want to encourage always using the kwargs for this or are fine with just using utter_message() for text without `text=`? If the latter, we can close

PR #138, opened 2020-01-02
Update Reminder* functions for intent-triggering

JEM-Mosig:
**Proposed changes**:
- Replace the `action` with an `intent` parameter for `ReminderScheduled` and add an `entity` parameter
- Replace the `action` with an `intent` parameter for `ReminderCancelled`

The release should be synced with the release of https://github.com/RasaHQ/rasa/pull/4964 in Rasa.

- [x] updated the changelog


rasa_sdk/events.py
```diff
@@ -68,16 +68,18 @@ def AllSlotsReset(timestamp: Optional[float] = None) -> EventType:
 
 # noinspection PyPep8Naming
 def ReminderScheduled(
-    action_name: Text,
+    intent: Text,
     trigger_date_time: datetime.datetime,
+    entities: Union[List[Dict[Text, Any]], Dict[Text, Text]] = None,
```
wochinge:
```suggestion
    entities: Optional[Union[List[Dict[Text, Any]], Dict[Text, Text]]] = None,
```

rasa_sdk/events.py
```diff
@@ -86,12 +88,16 @@ def ReminderScheduled(
 
 # noinspection PyPep8Naming
 def ReminderCancelled(
-    action_name: Text, name: Optional[Text] = None, timestamp: Optional[float] = None
+    name: Optional[Text] = None,
+    intent_name: Optional[Text] = None,
+    entities: Union[List[Dict[Text, Any]], Dict[Text, Text]] = None,
```
wochinge:
```suggestion
    entities: Optional[Union[List[Dict[Text, Any]], Dict[Text, Text]]] = None,
```

JEM-Mosig:
> Also is there a way that we can make this backwards compatible?
> E.g. by keeping the `action` parameter and returning the event in the old way in case `action` was given?

I don't think that's a good idea. If users want to use the old version of reminders, then they have to use the old version of Rasa anyway, and that should always be combined with the old version of the Rasa SDK. Don't you think?

rasa_sdk/events.py
```diff
@@ -86,12 +88,16 @@ def ReminderScheduled(
 
 # noinspection PyPep8Naming
 def ReminderCancelled(
-    action_name: Text, name: Optional[Text] = None, timestamp: Optional[float] = None
+    name: Optional[Text] = None,
+    intent_name: Optional[Text] = None,
+    entities: Optional[Union[List[Dict[Text, Any]], Dict[Text, Text]]] = None,
+    timestamp: Optional[float] = None,
 ) -> EventType:
```
wochinge: can we add a warning in case `intent_name` starts with `utter_` or `action_`? That's also sth we could unit test then.

JEM-Mosig: Done. I am not sure if it was correct to declare it as `DeprecationWarning`, though.

JEM-Mosig: Oh, and doing unit tests now...

wochinge: Should be a `FutureWarning`

JEM-Mosig: Ah, ok. I thought it's meant for the developers who develop the bot. Not for the enduser. So why is it `Future`?

JEM-Mosig: I've changed it now.

wochinge:
Mhm, actually doing a review of your own code can be quite useful.

rasa_sdk/events.py
```diff
@@ -68,16 +69,30 @@ def AllSlotsReset(timestamp: Optional[float] = None) -> EventType:
...
 def ReminderScheduled(
-    action_name: Text,
+    intent_name: Text,
     trigger_date_time: datetime.datetime,
+    entities: Optional[Union[List[Dict[Text, Any]], Dict[Text, Text]]] = None,
     name: Optional[Text] = None,
     kill_on_user_message: bool = True,
     timestamp: Optional[float] = None,
 ) -> EventType:
+    if intent_name.startswith("utter_"):
```
wochinge: optional: One warning for both cases is fine imo, but up to you.

tests/test_events.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+
+import warnings
+from datetime import datetime
+
+import rasa_sdk.events as events
+
+
+def test_reminder_scheduled_correctly():
+    with warnings.catch_warnings(record=True) as warning:
```
wochinge: can we please test the warnings as described here https://docs.pytest.org/en/latest/warnings.html#warns (`with pytest.warns(FutureWarning) ...`)

tests/test_events.py
```diff
@@ -0,0 +1,46 @@
...
+        events.ReminderScheduled("greet", datetime.now())
+        assert len(warning) == 0
+
+
+def test_reminder_scheduled_with_action():
```
wochinge: and please use `pytest.parametrize` for the tests, makes them a lot shorter :-)

JEM-Mosig:
> Mhm, actually doing a review of your own code can be quite useful.

Ah. I thought this was a malfunction, or I clicked something wrong, and I was confused because there was no way to 'cancel'. So I can compare my code to master there... yes, that's actually useful. Sorry, still getting used to GitHub.

rasa_sdk/events.py
```diff
@@ -86,12 +97,24 @@ def ReminderScheduled(
 
 # noinspection PyPep8Naming
 def ReminderCancelled(
-    action_name: Text, name: Optional[Text] = None, timestamp: Optional[float] = None
+    name: Optional[Text] = None,
+    intent_name: Optional[Text] = None,
+    entities: Optional[Union[List[Dict[Text, Any]], Dict[Text, Text]]] = None,
+    timestamp: Optional[float] = None,
 ) -> EventType:
+    if intent_name and (
```
wochinge: that's duplicate. can we put it in a function and re-use it?

tests/test_events.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from datetime import datetime
+
+import rasa_sdk.events as events
+
+
```
wochinge: 💯

rasa_sdk/events.py
JEM-Mosig: Is it ok to call the function `_is_probably_action_name`? I know it's not an "intent" like `_check_if_probably_action_name`, but in this case the former is actually more clear.

wochinge: good to me :-)

wochinge: wait, `is_...` indicates that the function returns a `bool` to me :-D Maybe something like `_check_potential_action_usage`?

JEM-Mosig: It does return `bool`. I just pushed it, for clarification.

JEM-Mosig: The messages are different for `ReminderScheduled` and `ReminderCancelled`, so I only "outsourced" the condition.

PR #163, opened 2020-03-31
Fill slot with a composite entity

tabergma:
**Proposed changes**:
Allow to fill a slot with an entity that has a specific role/group.
Example:
```
        def slot_mappings(self):
            return {
                "some_slot": self.from_entity(entity="some_entity", role="some_role")
            }
```

NOTE: Should not break anything, but functionality is only supported if branch `composite-entities` in Rasa Open Source is merged/released.


changelog/3765.feature.rst
```diff
@@ -0,0 +1,5 @@
+Add support for composite entities.
+
+If you use ``from_entity`` in your custom slot mapping, you can now also specify a role and group label.
+If you set a role or group label, the slot is only filled if the entity has the specific role or group label set.
```
erohmensing, line 4: And if you provide both, it has to match both right? (let me know if that use case doesn't even make sense)

rasa_sdk/forms.py
```diff
@@ -40,6 +40,8 @@ def from_entity(
         entity: Text,
         intent: Optional[Union[Text, List[Text]]] = None,
         not_intent: Optional[Union[Text, List[Text]]] = None,
+        role: Optional[Text] = None,
+        group: Optional[Text] = None,
```
erohmensing, line 44: Should these also optionally be lists? So that you could provide e.g. 2 out of 3 roles it could be?

tests/test_forms.py
```diff
@@ -170,6 +170,126 @@ def slot_mappings(self):
...
+    tracker = Tracker(
+        "default",
+        {"requested_slot": "some_slot"},
+        {
+            "intent": {"name": "some_intent", "confidence": 1.0},
+            "entities": [
+                {
+                    "entity": "some_entity",
+                    "value": "some_value",
+                    "role": "some_other_role",
```
erohmensing, line 197: Where if your other branch does this get added to the UserUttered event? (i took a short look but not too in depth)

tests/test_forms.py
```diff
@@ -170,6 +170,126 @@ def slot_mappings(self):
...
+        [],
+        False,
+        None,
+        {},
+        "action_listen",
+    )
+
+    slot_values = form.extract_requested_slot(CollectingDispatcher(), tracker, {})
+    # check that the value was extracted for correct intent
+    assert slot_values == {}
```
erohmensing, line 210: Please change this comment (and the other comments in the tests) to reflect what you're asserting

tests/test_forms.py
```diff
@@ -170,6 +170,126 @@ def slot_mappings(self):
...
+        {},
+        "action_listen",
+    )
+
+    slot_values = form.extract_requested_slot(CollectingDispatcher(), tracker, {})
+    # check that the value was not extracted for incorrect intent
+    assert slot_values == {"some_slot": "some_value"}
+
+
+def test_extract_requested_slot_from_entity_with_group():
```
erohmensing, line 233: We should test a combo of these, i guess, e.g. that if the group matches and the role doesn't, the slot is not filled

erohmensing: (well, i guess this goes in with my first comment -- not sure if this use case makes sense 😄

changelog/3765.feature.rst
tabergma: Yes, need to match both. This case is not very likely, but it might still happen.

rasa_sdk/forms.py
tabergma: That should not happen. You also need to decide for just ONE entity label. Same should happen for role and group.

erohmensing: That's true, I guess you could provide 2 mappings if for some reason you wanted to be able to pick up from both

tests/test_forms.py
tabergma: `UserUttered` has a list of entities (dict). Adding more values to the dict should be fine. They should be stored by default.

rasa_sdk/forms.py
tabergma: You mean in the slot mapping itself, right? So nothing I need to update.

erohmensing: nope, you should be able to provide a list i think

erohmensing: e.g.
```
return {
                "some_slot": [
                      self.from_entity(entity="some_entity", role="role1"),
                      self.from_entity(entity="some_entity", role="role2")
                ]
            }
```
this is how you do it if you wanted to have it pick up from one of 2 different entities

tabergma: ah ok, great.

tabergma:
I don't want to merge the PR yet, as I first want to get some user feedback. Will close this PR for now and reopen it once we decide that we actually want to merge the feature.

PR #164, opened 2020-04-01
Add auto-reload upon file timestamp change

federicotdn:
**Proposed changes**:
- Add a new CLI argument, `--auto-reload`. If enabled, modules containing actions will be reloaded when an HTTP request is received, but only if the files have been changed since they were last loaded.
- Closes https://github.com/RasaHQ/rasa/issues/4058.

Implementation details:

When we load a regular package (directory with `__init__.py` inside), or a file module (e.g. `actions.py`), the file path and last-modified timestamp is recorded. Whenever an HTTP request is received, before running the actions, all current file timestamps are checked against the recorded ones (there is no polling or file modification events involved). This process is quick because it only involves comparing floats. All files that have been modified since the last check are re-loaded using `importlib.reload`. After that, the whole `Action` class hierarchy is re-scanned, and the classes are registered.

One of the problems of re-loading modules is that the classes that have been already defined there aren't deleted or anything, so they are still part of the `Action` hierarchy. The consequence of this is that if a module containing `MyAction` is re-loaded, then there will be two `MyAction` classes in the hierarchy. To get around this, each time an action class is registered, it is "tagged" with `action_class._sdk_loaded = True`. That way, when we re-scan the `Action` class hierarchy, we know what actions we have already loaded in the past, and we can ignore them.

### DIFF
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 import types
-from typing import List, Text, Union, Optional, Any
+from typing import List, Text, Union, Optional, Any, NoReturn
 from ssl import SSLContext
 
 from sanic import Sanic, response
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -59,6 +59,7 @@ def create_argument_parser():
 def create_app(
     action_package_name: Union[Text, types.ModuleType],
     cors_origins: Union[Text, List[Text], None] = "*",
+    auto_reload: bool = False,
 ) -> Sanic:
     app = Sanic(__name__, configure_logging=False)
 
@@ -82,6 +83,10 @@ def create_app(
             return response.json(body, status=400)
 
         utils.check_version_compatibility(action_call.get("version"))
+
+        if auto_reload:
+            executor.reload()
+
         try:
             result = await executor.run(action_call)
         except ActionExecutionRejection as e:
@@ -98,6 +103,9 @@ def create_app(
     @app.get("/actions")
     async def actions(_) -> HTTPResponse:
         """List all registered actions."""
+        if auto_reload:
+            executor.reload()
+
         body = [{"name": k} for k in executor.actions.keys()]
         return response.json(body, status=200)
 
@@ -111,9 +119,12 @@ def run(
     ssl_certificate: Optional[Text] = None,
     ssl_keyfile: Optional[Text] = None,
     ssl_password: Optional[Text] = None,
+    auto_reload: bool = False,
 ):
     logger.info("Starting action endpoint server...")
-    app = create_app(action_package_name, cors_origins=cors_origins)
+    app = create_app(
+        action_package_name, cors_origins=cors_origins, auto_reload=auto_reload
+    )
     ssl_context = create_ssl_context(ssl_certificate, ssl_keyfile, ssl_password)
 
     app.run("0.0.0.0", port, ssl=ssl_context, workers=utils.number_of_sanic_workers())
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 import types
-from typing import List, Text, Union, Optional, Any, NoReturn
+from typing import List, Text, Union, Optional, Any
 from ssl import SSLContext
 
 from sanic import Sanic, response
@@ -131,7 +131,7 @@ def run(
     ssl_keyfile: Optional[Text] = None,
     ssl_password: Optional[Text] = None,
     auto_reload: bool = False,
-) -> NoReturn:
+) -> None:
     logger.info("Starting action endpoint server...")
     app = create_app(
         action_package_name, cors_origins=cors_origins, auto_reload=auto_reload
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -131,7 +131,7 @@ def run(
     ssl_keyfile: Optional[Text] = None,
     ssl_password: Optional[Text] = None,
     auto_reload: bool = False,
-):
+) -> NoReturn:
     logger.info("Starting action endpoint server...")
     app = create_app(
         action_package_name, cors_origins=cors_origins, auto_reload=auto_reload
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -61,6 +61,17 @@ def create_app(
     cors_origins: Union[Text, List[Text], None] = "*",
     auto_reload: bool = False,
 ) -> Sanic:
+    """Create a Sanic application and return it.
+
+    Args:
+        action_package_name: Name of the package or module to load actions
+            from.
+        cors_origins: CORS origins to allow.
+        auto_reload: When `True`, auto-reloading of actions is enabled.
+
+    Returns:
+        A new Sanic application ready to be run.
+    """
     app = Sanic(__name__, configure_logging=False)
 
     configure_cors(app, cors_origins)
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -193,12 +193,12 @@ class ActionExecutor:
     def _import_submodules(
         self, package: Union[Text, types.ModuleType], recursive: bool = True
     ) -> None:
-        """ Import all submodules of a module, recursively, including
-        subpackages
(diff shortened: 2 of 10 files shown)


rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
+            # package itself.
+            timestamp = os.path.getmtime(module_file)
+            self._modules[module_file] = (timestamp, module)
+
+        return module
+
     def register_package(self, package: Union[Text, types.ModuleType]) -> None:
-        from rasa_sdk.interfaces import Action
+        """Register all the `Action` subclasses contained in a Python module or
+        package.
```
wochinge, line 239: 🚀 Let's document that this function potentially kills everything in case loading the module goes wrong.

rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
         for action in actions:
-            meta = action.__dict__.get("Meta", False)
-            abstract = getattr(meta, "abstract", False)
             if (
                 not action.__module__.startswith("rasa_core.")
                 and not action.__module__.startswith("rasa.")
                 and not action.__module__.startswith("rasa_sdk.")
                 and not action.__module__.startswith("rasa_core_sdk.")
-                and not abstract
+                and not inspect.isabstract(action)
```
wochinge, line 265: how does this function work? :-D

wochinge: don't we need to do `Action(ABC)` then or use `@abstractmethod`??

rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
+        is larger than the previous one, the module is re-imported using
+        `importlib.reload`.
+
+        If one or more modules are reloaded during this process, the entire
+        `Action` class hierarchy is scanned again to see what new classes can
+        be registered.
+        """
+        to_reload = {}
+
+        for path, (timestamp, module) in self._modules.items():
```
wochinge, line 283: I'd split that method up in `_find_modules_to_reload` and `_reload_modules` and avoid having such a long method body.

rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
+            try:
+                new_timestamp = os.path.getmtime(path)
+            except OSError:
+                # Ignore missing files
+                continue
+
+            if new_timestamp > timestamp:
+                to_reload[path] = (new_timestamp, module)
+
+        module_reloaded = False
```
wochinge, line 293: ?
```suggestion
        all_modules_reloaded = False
```

setup.cfg
```diff
@@ -13,4 +13,4 @@ license_file = LICENSE.txt
 
 [flake8]
 max-line-length = 88
-ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, F821
+ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811
```
wochinge, line 16: Why this change?

tests/test_executor.py
```diff
@@ -1,7 +1,42 @@
...
+        return []
+"""
+
+
+def _write_action_file(
+    package_path: Text,
+    file_name: Text,
+    class_name: Text,
+    action_name: Text,
+    message: Text = None,
```
wochinge, line 31:
```suggestion
    message: Optional[Text] = None,
```

tests/test_executor.py
```diff
@@ -30,3 +65,135 @@ def test_deprecated_utter_elements():
         "image": None,
         "attachment": None,
     }
+
+
+@pytest.fixture()
+def package_path() -> None:
```
wochinge, line 71:
```suggestion
def package_path() -> Generator[Text, None, None]:
```

rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
         self, package: Union[Text, types.ModuleType], recursive: bool = True
     ) -> None:
-        """ Import all submodules of a module, recursively, including
-        subpackages
+        """Import a module, or a package and its submodules recursively.
 
-        :param package: package (name or actual module)
-        :type package: str | module
-        :rtype: dict[str, types.ModuleType]
+        Args:
```
wochinge, line 198: 🚀

rasa_sdk/executor.py
```diff
@@ -176,54 +182,135 @@ def register_function(self, name: Text, f: Callable) -> None:
...
+            except OSError:
+                # Ignore missing files
+                continue
+
+            if new_timestamp > timestamp:
+                to_reload[path] = (new_timestamp, module)
+
+        module_reloaded = False
+
+        for path, (timestamp, module) in to_reload.items():
```
wochinge, line 295: should we run this in the background using `sanic_app.add_task`?

rasa_sdk/executor.py
federicotdn: Yes! I changed it because that's the canonical way of making abstract classes in Python. I wasn't able to find why the previous method was being used.

rasa_sdk/executor.py
federicotdn: `any_modules_reloaded` is better I think, since we only care if at least one was reloaded

setup.cfg
federicotdn: That's so that we get warnings for using variables that are not defined, not sure why it was disabled since it's super useful.

rasa_sdk/executor.py
federicotdn: In that case the task would run after the HTTP request has been handled, so that modules would get reloaded after the request. This would demand users to make two requests instead of one to see the new changes, which I think would be unexpected for them.

federicotdn: I think, since this feature is for development use only and that checking for timestamps is a quick process (specially with small amounts of files), that reloading them before handling the requests won't be a problem.

m-vdb:
@federicotdn @wochinge I just ran into this "issue" myself and spent 30min trying to understand why my slots where not picked up by my form. I simply assumed that the action server would reload automatically (maybe because I am familiar with Django?).

I am wondering if we wouldn't want to make it the default behaviour (and an option to disable it). Are there drawbacks to that approach?
Maybe we also need to update the doc on that specific aspect: "you need to reload your action server if file A and B change"?

federicotdn:
@m-vdb From what I've seen, auto-reloading of files always seems to be a development-only option (for example in Flask and Sanic). I think it makes sense to have it disabled by default, because in one way or another it always implies that you're doing some additional processing to check if the files have been updated or not (via polling, OS-level file change watching, etc.). I agree that we should update the docs though, maybe we could add also that this new feature exists as well.

m-vdb:
ok makes sense re development, and re docs 👍

rasa_sdk/executor.py
wochinge: makes sense 👍

rasa_sdk/executor.py
wochinge: @federicotdn Did you do this one yet?

rasa_sdk/endpoint.py
```diff
@@ -111,9 +130,12 @@ def run(
     ssl_certificate: Optional[Text] = None,
     ssl_keyfile: Optional[Text] = None,
     ssl_password: Optional[Text] = None,
+    auto_reload: bool = False,
 ):
```
wochinge, line 134:
```suggestion
) -> NoReturn:
```

federicotdn:
Since it's for development I think we can leave it. Also I couldn't find a tidy way of unloading modules permanently in Python.

rasa_sdk/executor.py
federicotdn: What would you change here? The project itself does not have any abstract classes that I could find 🤔

wochinge: ah, true. `class Action` is filtered out by the other checks anyway

JulianGerhard21:
@federicotdn @m-vdb 

If I understand the implementation correctly, only the action classes that have already been loaded initially are reloaded, i.e. if I separate action classes into separate files in the "Actions" folder, for example, and make changes to them or add a new file, they are not loaded or reloaded.

Is this correct?

PR #165, opened 2020-04-02
Use custom mapping "from_entity" to fill other slots

tabergma:
**Proposed changes**:
Use the custom slot mapping 'from_entity' defined by the user to fill other slots.

Will be merged into `composite-entities`. We gonna collect some user feedback before we actually gonna merge this into master.

### DIFF
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -189,7 +189,7 @@ class FormAction(Action):
 
         return intent_not_blacklisted or intent in mapping_intents
 
-    def slot_is_desired(
+    def entity_is_desired(
         self, requested_slot_mapping: Dict[Text, Any], slot: Text, tracker: "Tracker"
     ) -> bool:
         """Check whether slot should be filled or not."""
@@ -197,8 +197,8 @@ class FormAction(Action):
         # slot name is equal to the entity type
         slot_equal_entity = slot == requested_slot_mapping.get("entity")
 
-        # use the custom slot mapping defined by the user to fill a slot based on
-        # a role or group
+        # use the custom slot mapping defined by the user to check whether we can
+        # fil a slot with an entity that has a role or group set
         slot_fulfils_entity_mapping = False
         if requested_slot_mapping.get("role") or requested_slot_mapping.get("group"):
             matching_values = self.get_entity_value(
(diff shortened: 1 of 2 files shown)


rasa_sdk/forms.py
```diff
@@ -242,8 +264,8 @@ def extract_other_slots(
                     # by entity with the same name
                     should_fill_entity_slot = (
                         other_slot_mapping["type"] == "from_entity"
-                        and other_slot_mapping.get("entity") == slot
                         and self.intent_is_desired(other_slot_mapping, tracker)
+                        and self.slot_is_desired(other_slot_mapping, slot, tracker)
```
erohmensing, line 268: This is a bit confusing to me -- the slot is always desired if it's empty. Should this be `entity_is_desired`? I.e. I guess what you are changing here is that even if the entity_type from the other mapping matches name of the slot, that does not necessarily mean that the entity is desired unless the group/role is right as well. Am i interpreting that correctly?

tabergma: Yeah, `entity_is_desired` is a better name. So either the slot name is equal to the entity name (normal auto filling) or the slot mapping triggers (to allow to auto-fill slots with entities that have a role/group).

rasa_sdk/forms.py
```diff
@@ -265,7 +265,7 @@ def extract_other_slots(
                     should_fill_entity_slot = (
```
erohmensing, line 263: The comment above this should i guess be sth like "check whether the slot should be filled by an entity in the input"

rasa_sdk/forms.py
```diff
@@ -189,16 +189,16 @@ def intent_is_desired(
 
         return intent_not_blacklisted or intent in mapping_intents
 
-    def slot_is_desired(
+    def entity_is_desired(
         self, requested_slot_mapping: Dict[Text, Any], slot: Text, tracker: "Tracker"
     ) -> bool:
         """Check whether slot should be filled or not."""
```
erohmensing, line 195:
```suggestion
        """Check whether slot should be filled by an entity in the input or not."""
```

rasa_sdk/forms.py
```diff
@@ -189,6 +189,26 @@ def intent_is_desired(
...
+    def entity_is_desired(
+        self, requested_slot_mapping: Dict[Text, Any], slot: Text, tracker: "Tracker"
+    ) -> bool:
+        """Check whether slot should be filled by an entity in the input or not."""
+
+        # slot name is equal to the entity type
+        slot_equal_entity = slot == requested_slot_mapping.get("entity")
+
+        # use the custom slot mapping 'from_entity' defined by the user to check
+        # whether we can fil a slot with an entity
```
erohmensing, line 201:
```suggestion
        # whether we can fill a slot with an entity
```

PR #205, opened 2020-04-30
bump version to 2.0a1

tmbo:
**Proposed changes**:
- bump version to 2.0a1


tmbo:
well, I mean there is 1.10.x so if we at some point need a 1.11.x we can branch of that branch

PR #207, opened 2020-04-30
updated typo in docs

nicht:
**Proposed changes**:
- Fixed typo in method documentation

### DIFF
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -1,7 +1,8 @@
 import copy
 import logging
-import typing
 from typing import Any, Dict, Iterator, List, Optional, Text
+
+from rasa_sdk.executor import CollectingDispatcher
 
 logger = logging.getLogger(__name__)
 
@@ -230,14 +231,17 @@ class Action:
         raise NotImplementedError("An action must implement a name")
 
     async def run(
-        self, dispatcher, tracker: Tracker, domain: Dict[Text, Any]
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: Dict[Text, Any],
     ) -> List[Dict[Text, Any]]:
         """Execute the side effects of this action.
 
         Args:
             dispatcher: the dispatcher which is used to
                 send messages back to the user. Use
-                ``dipatcher.utter_message()`` or any other
+                ``dispatcher.utter_message()`` or any other
                 ``rasa_sdk.executor.CollectingDispatcher``
                 method.
             tracker: the state tracker for the current
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -229,10 +229,7 @@ class Action:
         raise NotImplementedError("An action must implement a name")
 
     async def run(
-        self,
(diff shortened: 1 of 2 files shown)


rasa_sdk/interfaces.py
```diff
@@ -230,14 +231,17 @@ def name(self) -> Text:
...
     ) -> List[Dict[Text, Any]]:
         """Execute the side effects of this action.
 
         Args:
             dispatcher: the dispatcher which is used to
                 send messages back to the user. Use
-                ``dipatcher.utter_message()`` or any other
+                ``dispatcher.utter_message()`` or any other
                 ``rasa_sdk.executor.CollectingDispatcher``
                 method.
```
erohmensing, line 246: Since the other methods are now deprecated, would you mind updating this to say that the dispatcher sends messages using the `dispatcher.utter_message()` method?

nicht: Sure! I will update this later and fix the tests

erohmensing: Great! Am actually wondering why the tests are breaking, this doesn't seem like it should break anything. Perhaps it's an import loop...

nicht: Yeah, that seems right. This is the output i get after running the tests locally 

image

rasa_sdk/interfaces.py
```diff
@@ -230,14 +231,17 @@ def name(self) -> Text:
         raise NotImplementedError("An action must implement a name")
 
     async def run(
-        self, dispatcher, tracker: Tracker, domain: Dict[Text, Any]
+        self,
+        dispatcher: CollectingDispatcher,
```
nicht, line 235: @erohmensing, we have a problem with the definition of this class with parameter typing, I only realized now by reviewing the code and running the tests... It is causing a circular dependency between the `interfaces.py` and` executor.py` files. Do you think it pays to just change the docs for that method and not add the typing?

sara-tagger:
Thanks for submitting a pull request 🚀 @alwx will take a look at it as soon as possible ✨

rasa_sdk/interfaces.py
erohmensing: Yes, seems so. I can fix the tests by adding an `if typing.TYPE_CHECKING` check for the import, but then pytype ultimately fails. Let's still update the typo though.

rasa_sdk/interfaces.py
```diff
@@ -230,16 +231,17 @@ def name(self) -> Text:
...
     ) -> List[Dict[Text, Any]]:
         """Execute the side effects of this action.
 
         Args:
             dispatcher: the dispatcher which is used to
                 send messages back to the user. Use
-                ``dipatcher.utter_message()`` or any other
-                ``rasa_sdk.executor.CollectingDispatcher``
-                method.
+                ``dispatcher.utter_message()`` for sending messages
```
erohmensing, line 244:
```suggestion
                ``dispatcher.utter_message()`` for sending messages.
```

erohmensing:
@nicht can you remove the typing as discussed and then we can move forward with fixing the typo?

PR #236, opened 2020-07-12
Proper Abstract Action Check

### LINKED ISSUE #233: Creating subclass of subclass of Action works incorrectly
**Rasa version**: 1.10

**Rasa SDK version**: 1.10

**Python version**: 3.7

**Operating system** (windows, osx, ...): Linux

**Issue**: Creating subclass `B` of subclass `A` of class `Action` would lead to registering of class `A`, but not registering class `B`, unless class `A` is abstract. This is because in `register_action` method of `ActionExecutor` class is getting tagged by setting attribute `_sdk_loaded`. The case is that it will be set in class `A`, and since `B` is subclass of `A`, `B` will be skipped since it already has this attribute.

Class `A` registered, `B` is not
```

```

**Content of configuration file (config.yml)** (if relevant):
```yml

```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

**Contents of actions.py** (if  relevant):
```python
class A(Action):
    pass
class B(A):
    pass
```

lhr0909:
~This change tries to resolve issue 233 , since the `inpsect.isabstract()` only works on instantiated class instances.~

Edit: As I was working on this by adding unit test, I found that the original code also passes the test, so there is no need for the code change. I am going to commit just the unit test to cover the abstract action register, so hopefully it is still usefull.

**Proposed changes**:
~- in the executor's `_register_all_actions()` method, we will try to instantiate the action class and make sure it is not abstract before we register.~
- Added unit test to cover the case with the abstract class check in ActionExecutor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,7 @@ venv.bak/
 # mypy
 .mypy_cache/
 .idea/
+.vscode/
 examples/moodbot/models/
 
 # emacs
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -261,14 +261,27 @@ class ActionExecutor:
         actions = utils.all_subclasses(Action)
 
         for action in actions:
+            logger.debug(f"checking action {action.__module__}")
+
             if (
-                not action.__module__.startswith("rasa_core.")
-                and not action.__module__.startswith("rasa.")
-                and not action.__module__.startswith("rasa_sdk.")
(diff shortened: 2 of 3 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @koaning will take a look at it as soon as possible ✨

lhr0909:
I will be adding proper tests in the coming days

lhr0909:
Closing since I think I made a mistake when I was running my test.

Issue #80, opened 2019-05-14
add activation function in forms


erohmensing:
The idea is that we would add an activation function where any actions that need to be performed before the form starts go. Then we would maybe also want to reinforce that forms should be called directly after inputs (e.g. by checking stories and tellling people to put anything that came before the form action in their activation function to avoid undesired behavior like slots from activating input not filling)

wochinge:
@erohmensing What would these actions do?

erohmensing:
they could do even simple things such as utterances -- the main issue is that if you add a different action before the form action, requested slots filled from the user input (that aren't auto-filled into slots) are not validated correctly. i was of the opinion that that was a bug, but @Ghostvv suggested doing it this way instead. thread: https://github.com/RasaHQ/rasa/issues/3279#issuecomment-485684005

i.e. minimum reproducible: something like
```
*subscribe_newsletter{"email":"ella@ella.ella"} 
          - utter_can_do
          - subscribe_newsletter_form
```
with `autofill:false` and a validation method that e.g. says "reject it if it includes `ella`" won't reject it

yangjianxin1:
I can not agree more with you，when I ask "What is the weather like in Beijing"，I prefer that the chatbot would fill the slot "weather_city" immediately，rather than doing nothing。This question has confused me for a long time。Now I know that we can rewrite _activate_if_required() to implement this goal.

yangjianxin1:
@erohmensing I can not agree more with you，when I ask "What is the weather like in Beijing"，I prefer that the chatbot would fill the slot "weather_city" immediately，rather than doing nothing。This question has confused me for a long time。Now I know that we can rewrite _activate_if_required() to implement this goal.

wochinge:
I deprio this for now since their might be bigger changes to forms in the near future

wochinge:
@Ghostvv 

> the main issue is that if you add a different action before the form action, requested slots filled from the user input (that aren't auto-filled into slots) are not validated correctly.

This is no longer true and we can close it, correct?

Ghostvv:
not for sdk, one line needs to be changed there

PR #239, opened 2020-07-20
allow validating user input on activation even if there are actions in between input and form

Ghostvv:
**Proposed changes**:
- add that form needs to validate if active_form dict is empty, because it means it is first form call
- fixes https://github.com/RasaHQ/rasa-sdk/issues/80

### DIFF
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -543,9 +543,12 @@ class FormAction(Action):
             - the form is called after `action_listen`
             - form validation was not cancelled
         """
-        if tracker.latest_action_name == "action_listen" and tracker.active_form.get(
-            "validate", True
-        ):
+        # no active_form means that it is called during activation
+        need_validation = not tracker.active_form or (
+            tracker.latest_action_name == "action_listen"
+            and tracker.active_form.get("validate", True)
+        )
+        if need_validation:
             logger.debug(f"Validating user input '{tracker.latest_message}'")
             if utils.is_coroutine_action(self.validate):
                 return await self.validate(dispatcher, tracker, domain)
--- a/changelog/239.improvement.rst
+++ b/changelog/239.improvement.rst
@@ -1,0 +1,1 @@
+Validate user input on activation even if there is an action in between.
--- a/changelog/239.improvement.rst
+++ b/changelog/239.improvement.rst
@@ -1,1 +1,1 @@
-Validate user input on activation even if there is an action in between.
+Validate user input during form activation even if there is an action in between.
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -33,7 +33,7 @@ class Tracker:
         events: List[Dict[Text, Any]],
         paused: bool,
         followup_action: Optional[Text],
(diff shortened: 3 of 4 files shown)


rasa_sdk/forms.py
```diff
@@ -543,9 +543,12 @@ async def _validate_if_required(
             - the form is called after `action_listen`
             - form validation was not cancelled
         """
-        if tracker.latest_action_name == "action_listen" and tracker.active_form.get(
-            "validate", True
-        ):
+        # no active_form means that it is called during activation
+        need_validation = not tracker.active_form or (
+            tracker.latest_action_name == "action_listen"
```
wochinge, line 548: let's use the constant `ACTION_LISTEN_NAME` and make sure it's covered by a test

Ghostvv: the whole `FormAction` uses strings, it'll be weird to use constant only here, and I don't think it makes sense to refactor it, since we're going to deprecate it.

wochinge: > the whole FormAction uses strings, it'll be weird to use constant only her

I strongly disagree. Just because something wasn't done correctly in the past, we shouldn't carry on doing this. 

> and I don't think it makes sense to refactor it, since we're going to deprecate it.

It'll still stay in the code for a while, so why not doing the 10sec code change and be done with it?

wochinge: >  it is user facing code, and constant is way more confusing then string, I'd keep the string here

We already have an existing constant for this in the SDK and we should use it. It might make sense to not use a constant (or delay its introduction) if there is no constant yet and the raw string only appears in very few places. As we already have the constant I see using the constant as a must do.
How is `ACTION_LISTEN_NAME` more confusing than `"action_listen"`?

Ghostvv: > We already have an existing constant for this in the SDK and we should use it.

this constant is inside `interfaces.py`, quite hard for users to find.

> How is ACTION_LISTEN_NAME more confusing than "action_listen"?

it is super not clear what it is, if you are not familiar with rasa systems, while the string explicitly states that it is some action

wochinge: > this constant is inside interfaces.py, quite hard for users to find.

The import makes it clear where to find + the users shouldn't change this part anyway. 

> it is super not clear what it is, if you are not familiar with rasa systems, while the string explicitly states that it is some action

Both, string and constant, contain the `action_` part so both are clear. 

@degiz @erohmensing @alwx What do you think? I feel this discussion is a bit stuck 😄

degiz: I'm for using the "ACTION_LISTEN_NAME" here for mentioned reasons:
- I agree that even though we deprecate that it's a tiny change that brings more benefits
- a good habit to not use explicit strings, and this is a good chance to exercise that. If the string has to change at some point of time it's good to have only one constant, not to search for it
- the source file is not being read as a book, users have IDEs or text editors or `grep`

Ghostvv: in this particular case the source is used as a book because users look at it in GitHub browser and not in IDE

Issue #245, opened 2020-08-17
Switch to mypy


wochinge:
**Problem:**
We repeatedly had code issues which could have code by type checks. We currently use `pytype` and it didn't find very simple things like the one which @m-vdb shared [here](https://gist.github.com/m-vdb/d24a45120cae0daa886f35e2aa032260) and could have prevented [this issue](https://github.com/RasaHQ/rasa-x/pull/3328).

**Proposed Solution**
`mypy` seems better suited and would have caught these issues. To gain more experience with it, we should start switching our smallest project, `rasa-sdk` to `mypy`. Depending on the results we can decide whether it's worth switching `rasa` or `rasa-`x, too.

**Steps**
- [ ] Switch `rasa-sdk` to `mypy`
- [ ] Make notes of all points of frictions / cool things during the process
- [ ] Compare the speed of the type checks (how much faster / slower is it than `pytype`

nathan-chappell:
glad to hear you guys are switching to mypy.  I've been using it for awhile, and am looking forward to using it with rasa.

m-vdb:
@nathan-chappell thanks for your feedback 🎉 `mypy` is now in use in `rasa`'s master (as of https://github.com/RasaHQ/rasa/pull/6470). Enjoy!

PR #247, opened 2020-08-21
Switch to mypy

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa-sdk/issues/245
- on master, with `pytype` it took 35s; now it seems to take 2s with `mypy`


rasa_sdk/types.py
```diff
@@ -0,0 +1,33 @@
...
+class DomainDict(TypedDict):
+    intents: List[Union[Text, Dict[Text, Any]]]
+    entities: List[Text]
+    slots: Dict[Any, Dict[Text, Any]]
+    responses: Dict[Text, List[Dict[Text, Any]]]
+    actions: List[Text]
+    forms: List[Union[Text, Dict]]
+
+
+class ActionCall(TypedDict):
```
m-vdb, line 28: these type definitions need proper review + docstrings

setup.cfg
```diff
@@ -14,3 +14,6 @@ license_file = LICENSE.txt
 [flake8]
 max-line-length = 88
 ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811
+
+[mypy]
+ignore_missing_imports = True
```
m-vdb, line 19: this ignores missing type definitions for 3rd party modules

rasa_sdk/forms.py
```diff
@@ -213,17 +214,18 @@ def entity_is_desired(
         """
 
         # slot name is equal to the entity type
-        other_slot_equals_entity = other_slot == other_slot_mapping.get("entity")
+        entity_type = other_slot_mapping.get("entity")
+        other_slot_equals_entity = other_slot == entity_type
```
m-vdb, line 218: this is part of a small refactoring done in 4dec28a

rasa_sdk/forms.py
```diff
@@ -567,7 +578,10 @@ async def _validate_if_required(
             if utils.is_coroutine_action(self.validate):
                 return await self.validate(dispatcher, tracker, domain)
             else:
-                return self.validate(dispatcher, tracker, domain)
+                # see https://github.com/python/mypy/issues/5206
```
m-vdb, line 581: if we have to do this a lot in rasa, we might consider contributing to mypy to support useful type guards

rasa_sdk/knowledge_base/actions.py
```diff
@@ -45,7 +50,7 @@ def utter_attribute_value(
         object_name: Text,
         attribute_name: Text,
         attribute_value: Text,
-    ) -> None:
```
tmbo, line 48: should we remove these? as far as I remember the pep suggested to keep them, but not sure

rasa_sdk/types.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, Dict, List, Optional, Text, Union
+
+from typing_extensions import TypedDict
+
+
+class TrackerState(TypedDict):
```
tmbo, line 6: while you are at it, great opportunity to add comments to the attributes :P

tmbo, line 6: should we rather use pythons new data classes? (and a backport for 3.6 https://pypi.org/project/dataclasses/)

tmbo (commented):
looks great, excited to replace pytype 💀  😂

rasa_sdk/knowledge_base/actions.py
m-vdb: The only thing that I could find in PEP regarding `None` return values is for the `__init__()` method, where it's considered a best practice to return `None`, otherwise the type checker might not validate the types for the method.

The reason `mypy` complains is because of this: https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-called-function-returns-a-value-func-returns-value

I would agree that it's better to make a clear distinction between a function that is supposed to return something (and you're suppose to do something with the returned value), versus a function that doesn't return anything (and you shouldn't do anything with the return value).

rasa_sdk/types.py
m-vdb: I think that could be better indeed, but I do not know enough about the specifics of the tracker state / action call:
- would renaming the method `from_dict()` to `from_dataclass()` be acceptable?
- is the type definition I made here valid? (I marked some elements required, before they were optional)

m-vdb: after thinking a bit more about it, I'm not sure that we would benefit from the dataclass here (vs keeping the TypedDict). And if we wanted to do this it might change the public API of the SDK. For instance, instead of calling custom action with plain dicts, we would call them with dataclasses. And I'm not sure it's worth the trouble. WDYT?

rasa_sdk/forms.py
tmbo: yes this sounds like a good idea :+1: maybe something for someone from the team to contribute?

rasa_sdk/types.py
tmbo: ok that sounds reasonable, let's stick with the typeddicts :+1:

rasa_sdk/types.py
```diff
@@ -0,0 +1,58 @@
...
+    # the name of the previously executed action or text of e2e action
+    latest_action_name: Optional[Text]
+
+
+class DomainDict(TypedDict):
+    """
+    A dictionary representation of the domain.
+    """
+
+    intents: List[Union[Text, Dict[Text, Any]]]
```
tmbo, line 36: can this really contain strings?

tmbo (approved):
looks great - amazing push for better tooling :rocket:

rasa_sdk/forms.py
m-vdb: yes definitely. I will keep this as a note for now, and if this comes back I'll mention it (in backend eng meeting for instance)

rasa_sdk/types.py
m-vdb: no you're right. The domain is implemented to handle both, but the dictionary representation of the domain only contains list of dicts in the `intents` key (cf. [here](https://github.com/RasaHQ/rasa/blob/05be168f7d3a68c64453b3c444cf4dff1eea1ce4/rasa/shared/core/domain.py#L825-L853))

m-vdb: and it's the same for forms actually

PR #248, opened 2020-08-24
remove pii from non-debug logs

erohmensing:
**Proposed changes**:
- Remove any non-debug logs that could feasibly contain PII. 
- Turn `Failed to extract slot email with action subscribe_newsletter_form` into a debug log because it is expected behavior for unexpected input, and clarify that in the log to make it less confusing.

### DIFF
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -101,7 +101,7 @@ def create_app(
         try:
             result = await executor.run(action_call)
         except ActionExecutionRejection as e:
-            logger.error(e)
+            logger.debug(e)
             body = {"error": e.message, "action_name": e.action_name}
             return response.json(body, status=400)
         except ActionNotFoundException as e:
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -336,18 +336,21 @@ class ActionExecutor:
     @staticmethod
     def validate_events(events: List[Dict[Text, Any]], action_name: Text):
         validated = []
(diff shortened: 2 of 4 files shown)


akelad:
i'm not sure i am the right person to review this - did you talk to tom or sasha about whether we want to do it this way? with the `logger.debug(e)` i mean

erohmensing:
Waiting for his input on the thread. In any case the one I changed should actually be a debug log, it doesn't make sense for it to be an error (see description). but the question does affect L108 there

rasa_sdk/endpoint.py
```diff
@@ -101,7 +101,7 @@ async def webhook(request: Request) -> HTTPResponse:
         try:
             result = await executor.run(action_call)
         except ActionExecutionRejection as e:
-            logger.error(e)
+            logger.debug(e)
```
erohmensing, line 104: This one got changed to debug not because it contained PII (the only ones we raise contain only action names and slot names) but because it is expected behavior (the form rejects when you give it unexpected input). i clarified the response body message to reflect that, but that is why we are not raising an error.

degiz (approved):
Cool 👍 
@erohmensing @akelad since that (PII) becomes a hot topic and most probably wanted by enterprises, should we create a separate issue to go over all repos and do the similar changes as this one?

erohmensing:
I was going to keep track of it in [this issue!](https://github.com/RasaHQ/rasa-x/issues/3474)

PR #260, opened 2020-09-01
Switch to pyright

m-vdb:
## Comparison with `mypy`

- Not really granular error codes (`reportGeneralTypeIssues` looks like a catch-all), even though there are different error codes in the JSON config, and the message for each of the errors are different
- Documentation is in GitHub md files (vs readthedocs site for `mypy`)
- doesn't support type guards (same as `mypy`)
- performance is similar on rasa-sdk, around 2 seconds to run
- Additional setup for someone who wants to run the type checks of the repo (they would need to setup `yarn` or `npm` instead of simply using python with `mypy`). Not sure if this is a blocker or not, it simply seems weird to me from a contributor experience POV
- Repository seems a bit less active than `mypy` (less issues and PRs, less stars, more code frequency)
- `pyright` seems more recent (~1 year) than `mypy` (~7 years). It's backed by Microsoft whereas `mypy` seems to be backed mainly by an open source community
- Editor supports: [`pyright`](https://github.com/microsoft/pyright#installation) and [`mypy`](https://github.com/python/mypy#ide-linter-integrations-and-pre-commit) seem comparable.


m-vdb:
@tmbo @gausie here is a draft PR of how the setup for `pyright` would look like. I did a comparison in the PR description. To me, the benefits from `pyright` over `mypy` are not clear, and it would add friction for contributors to a python codebase. Happy to hear your thoughts on this!

gausie:
If there's no clear benefit, it seems like we shouldn't take on pyright! As I said on slack i found it to be vastly better in terms of type inference. Suggest we close and continue with mypy!

m-vdb:
Thanks for your input Sam 😄

PR #288, opened 2020-10-01
FormSlotsValidatorAction added

alwx:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6819

### DIFF
--- a/changelog/288.improvement.md
+++ b/changelog/288.improvement.md
@@ -1,2 +1,22 @@
 Add the `FormSlotsValidatorAction` abstract class that can be used
 to validate slots which were extracted by a Form.
+
+Example:
+
+```python
+class MyFormValidationAction(FormValidationAction):
+    def name(self) -> Text:
+        return "some_form"
+
+    def validate_slot1(
+        self,
+        slot_value: Any,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Dict[Text, Any]:
+        if slot_value == "correct_value":
+            return {
+                "slot1": "validated_value",
+            }
+```
--- a/changelog/288.improvement.md
+++ b/changelog/288.improvement.md
@@ -19,4 +19,7 @@ class MyFormValidationAction(FormValidationAction):
             return {
                 "slot1": "validated_value",
             }
+        return {
+            "slot1": None,
+        }
 ```
--- a/changelog/288.improvement.md
+++ b/changelog/288.improvement.md
@@ -1,0 +1,2 @@
+Add the `FormSlotsValidatorAction` abstract class that can be used
+to validate slots which were extracted by a Form.
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -711,7 +711,7 @@ class FormValidationAction(Action):
             )
                 return slot_value == "correct_value"
     ```
-        """
+    """
 
     def name(self) -> Text:
         """Unique identifier of this action.
@@ -750,7 +750,9 @@ class FormValidationAction(Action):
                     slot_value, dispatcher, tracker, domain
                 )
             else:
-                validation_output = validate_func(slot_value, dispatcher, tracker, domain)
+                validation_output = validate_func(
+                    slot_value, dispatcher, tracker, domain
+                )
 
             if validation_output:
                 validated_events.append(SlotSet(slot_name, slot_value))
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -691,3 +691,45 @@ class FormAction(Action):
                 return None
 
         return entity_type
+
+
+class FormSlotsValidatorAction(Action):
+    def name(self) -> Text:
+        """Unique identifier of this action."""
+
+        raise NotImplementedError("An action must implement a name")
+
+    def run(
+        self, dispatcher, tracker: Tracker, domain: Dict
+    ) -> List[EventType]:
+        """Execute the side effects of this action.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user. Use
+                ``dispatcher.utter_message()`` for sending messages.
+            tracker: the state tracker for the current
+                user. You can access slot values using
+                ``tracker.get_slot(slot_name)``, the most recent user message
+                is ``tracker.latest_message.text`` and any other
+                ``rasa_sdk.Tracker`` property.
+            domain: the bot's domain
+        Returns:
+            A dictionary of ``rasa_sdk.events.Event`` instances that is
+                returned through the endpoint
+        """
+
+        slots_to_validate: Dict[Text, Any] = tracker.form_slots_to_validate()
+        validation_events = []
+
+        for slot_name, slot_value in slots_to_validate.items():
+            function_name = f"validate_{slot_name}"
+            fn = getattr(self, function_name)
+            if fn(tracker, domain, slot_value):
+                validation_events.append(SlotSet(slot_name, slot_value))
+            else:
+                # Return a `SlotSet` event with value `None` to indicate that this
+                # slot still needs to be filled.
+                validation_events.append(SlotSet(slot_name, None))
+
+        return validation_events
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -693,13 +693,76 @@ class FormAction(Action):
         return entity_type
 
 
-class FormSlotsValidatorAction(Action):
-    """An action that validates if every extracted slot is valid."""
+class FormValidationAction(Action):
+    """An action that validates if every extracted slot is valid.
+
+    Example of usage:
+    ```
+        class MyFormValidationAction(FormValidationAction):
+            def name(self) -> Text:
+                return "some_form"
+
+            def validate_slot1(
+                self,
+                slot_value: Any,
+                dispatcher: "CollectingDispatcher",
+                tracker: "Tracker",
+                domain: "DomainDict",
+            )
+                return slot_value == "correct_value"
+    ```
+        """
 
     def name(self) -> Text:
-        """Unique identifier of this action."""
+        """Unique identifier of this action.
+
+        Returns:
+            Name of the action.
+        """
 
         raise NotImplementedError("An action must implement a name")
+
+    async def validate(
+        self,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> List[EventType]:
+        """Validate slots by calling a validation function for each slot.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            A dictionary of `rasa_sdk.events.Event` instances.
+        """
+        slots_to_validate: Dict[Text, Any] = tracker.form_slots_to_validate()
+        validated_events = []
+
+        for slot_name, slot_value in slots_to_validate.items():
+            function_name = f"validate_{slot_name}"
+            validate_func = getattr(self, function_name, lambda *x: False)
+
+            if utils.is_coroutine_action(validate_func):
+                validation_output = await validate_func(
+                    slot_value, dispatcher, tracker, domain
+                )
+            else:
+                validation_output = validate_func(slot_value, dispatcher, tracker, domain)
+
+            if validation_output:
+                validated_events.append(SlotSet(slot_name, slot_value))
+            else:
+                # Return a `SlotSet` event with value `None` to indicate that this
+                # slot still needs to be filled.
+                warnings.warn(
+                    f"Cannot validate `{slot_name}`: make sure the validation function is specified and "
+                    f"returns `True`"
+                )
+
+        return validated_events
 
     async def run(
         self,
@@ -715,20 +778,7 @@ class FormSlotsValidatorAction(Action):
             tracker: the state tracker for the current user.
             domain: the bot's domain.
         Returns:
-            A dictionary of ``rasa_sdk.events.Event`` instances.
+            A dictionary of `rasa_sdk.events.Event` instances.
         """
 
-        slots_to_validate: Dict[Text, Any] = tracker.form_slots_to_validate()
-        validation_events = []
-
-        for slot_name, slot_value in slots_to_validate.items():
-            function_name = f"validate_{slot_name}"
-            fn = getattr(self, function_name)
-            if fn(tracker, domain, slot_value):
-                validation_events.append(SlotSet(slot_name, slot_value))
-            else:
-                # Return a `SlotSet` event with value `None` to indicate that this
-                # slot still needs to be filled.
-                validation_events.append(SlotSet(slot_name, None))
-
-        return validation_events
+        return await self.validate(dispatcher, tracker, domain)
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -700,7 +700,9 @@ class FormSlotsValidatorAction(Action):
 
         raise NotImplementedError("An action must implement a name")
 
-    async def run(self, dispatcher, tracker: "Tracker", domain: Dict) -> List[EventType]:
+    async def run(
+        self, dispatcher, tracker: "Tracker", domain: Dict
+    ) -> List[EventType]:
         """Execute the side effects of this action.
 
         Args:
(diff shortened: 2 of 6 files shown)


alwx:
Will update https://rasa.com/docs/rasa/forms#validating-form-input once we merge this one :)

wochinge:
I'll try to squeeze in the review between retro and tech onboarding .Do you already have a matching docs PR on Rasa Open Source side?

rasa_sdk/forms.py
```diff
@@ -689,3 +689,44 @@ def _get_entity_type_of_slot_to_fill(
...
+            A dictionary of ``rasa_sdk.events.Event`` instances.
+        """
+
+        slots_to_validate: Dict[Text, Any] = tracker.form_slots_to_validate()
+        validation_events = []
+
+        for slot_name, slot_value in slots_to_validate.items():
+            function_name = f"validate_{slot_name}"
+            fn = getattr(self, function_name)
+            if fn(tracker, domain, slot_value):
```
wochinge, line 725: I think we should keep this in line to the behavior in the old deprecated `FormAction` (see [here](https://github.com/RasaHQ/rasa-sdk/blob/2fff067d26b64aaabaa5ce60163081ffe0418891/rasa_sdk/forms.py#L408)). This would also allow users to simply copy over their validate functions.

The signature should imo look like:

```python
def validate_my_slot(
        self,
        slot_dict: Dict[Text, Any],
        dispatcher: "CollectingDispatcher",
        tracker: "Tracker",
        domain: "DomainDict",
    ) -> List[EventType]
```

wochinge, line 725: also note that the validation methods might be `async` methods

rasa_sdk/forms.py
```diff
@@ -689,3 +689,44 @@ def _get_entity_type_of_slot_to_fill(
...
+                validation_events.append(SlotSet(slot_name, slot_value))
+            else:
+                # Return a `SlotSet` event with value `None` to indicate that this
```
wochinge, line 728: I think we should also print a warning for this. To be honest I'd just skip the validation for this slot then instead of marking it as invalid.

rasa_sdk/interfaces.py
```diff
@@ -250,6 +250,7 @@ def form_slots_to_validate(self) -> Dict[Text, Any]:
         if not self.active_loop:
             return slots_to_validate
 
+        print(self.events)
```
wochinge, line 253:
```suggestion
```

rasa_sdk/forms.py
wochinge, line 725: we don't provide a default value to `getattr` so this will always raise. Means we don't need to check `if fn` as `fn` will always be set. However, I think not raising, printing a warning and not validating the slot is the better option.

rasa_sdk/forms.py
```diff
@@ -689,3 +689,44 @@ def _get_entity_type_of_slot_to_fill(
...
+    ) -> List[EventType]:
+        """Execute the side effects of this action.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            A dictionary of ``rasa_sdk.events.Event`` instances.
```
wochinge, line 716:
```suggestion
            A dictionary of `rasa_sdk.events.Event` instances.
```

wochinge (commented):
Nice testing 🥇 Left a couple of comments 🙈

rasa_sdk/forms.py
```diff
@@ -689,3 +689,44 @@ def _get_entity_type_of_slot_to_fill(
...
+
+        for slot_name, slot_value in slots_to_validate.items():
+            function_name = f"validate_{slot_name}"
+            fn = getattr(self, function_name)
+            if fn(tracker, domain, slot_value):
+                validation_events.append(SlotSet(slot_name, slot_value))
+            else:
+                # Return a `SlotSet` event with value `None` to indicate that this
+                # slot still needs to be filled.
+                validation_events.append(SlotSet(slot_name, None))
```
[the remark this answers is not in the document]
alwx: Yep, makes sense!

rasa_sdk/forms.py
alwx: Not sure why you pasted this signature while the previous signature was clearly different :)
In addition, it seems like having `slot_dict` as an argument is useless because you can easily get slots from `tracker`.

changelog/288.improvement.md
```diff
@@ -0,0 +1,2 @@
+Add the `FormSlotsValidatorAction` abstract class that can be used
+to validate slots which were extracted by a Form.
```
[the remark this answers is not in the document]
wochinge: I meant actually here in the changelog. I think users are more likely to look into the changelog then the codebase.

rasa_sdk/forms.py
```diff
@@ -689,3 +689,96 @@ def _get_entity_type_of_slot_to_fill(
...
+                )
+
+            if validation_output:
+                validated_events.append(SlotSet(slot_name, slot_value))
+            else:
+                # Return a `SlotSet` event with value `None` to indicate that this
+                # slot still needs to be filled.
+                warnings.warn(
+                    f"Cannot validate `{slot_name}`: make sure the validation function is specified and "
+                    f"returns `True`"
```
wochinge, line 762: it shouldn't return `True` but a list of events, right?

tests/test_forms.py
```diff
@@ -1493,3 +1494,76 @@ async def test_submit(form_class: Type[FormAction]):
...
+            SlotSet("slot2", "incorrect_value"),
+            SlotSet("slot3", "some_value"),
+        ],
+        False,
+        None,
+        {"name": "some_form", "validate": True, "rejected": False},
+        "action_listen",
+    )
+
+    dispatcher = CollectingDispatcher()
```
wochinge, line 1564: let's also test the warning here

wochinge (commented):
looks great

rasa_sdk/forms.py
wochinge: sorry, overlooked the `slot_dict`. You're right, should be the extracted slot value instead.

rasa_sdk/forms.py
```diff
@@ -689,3 +689,96 @@ def _get_entity_type_of_slot_to_fill(
...
+            def name(self) -> Text:
+                return "some_form"
+
+            def validate_slot1(
+                self,
+                slot_value: Any,
+                dispatcher: "CollectingDispatcher",
+                tracker: "Tracker",
+                domain: "DomainDict",
+            ) -> bool:
```
wochinge, line 709: that should return a list of events, right?

rasa_sdk/forms.py
```diff
@@ -689,3 +689,96 @@ def _get_entity_type_of_slot_to_fill(
...
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            A dictionary of `rasa_sdk.events.Event` instances.
+        """
+        slots_to_validate: Dict[Text, Any] = tracker.form_slots_to_validate()
+        validated_events = []
```
wochinge, line 740: ?
```suggestion
        validated_slots = []
```

tests/test_forms.py
```diff
@@ -1493,3 +1494,76 @@ async def test_submit(form_class: Type[FormAction]):
 def test_form_deprecation():
     with pytest.warns(FutureWarning):
         FormAction()
+
+
+class TestFormValidationAction(FormValidationAction):
+    def name(self) -> Text:
+        return "some_form"
+
+    def validate_slot1(
```
wochinge, line 1503: can we make one `async` to test this as well?

alwx:
I made it work the way it worked before (in the current implementation events get updated in `slots` dictionary) but I still have no idea if it's the right behavior or not because there is nothing about it in the description.
@wochinge could you please check and verify if the logic is correct?

rasa_sdk/forms.py
```diff
@@ -689,3 +689,76 @@ def _get_entity_type_of_slot_to_fill(
...
+    """An action that validates if every extracted slot is valid."""
+
+    def name(self) -> Text:
+        """Unique identifier of this action.
+
+        Returns:
+            Name of the action.
+        """
+
+        raise NotImplementedError("An action must implement a name")
```
wochinge, line 704: we don't have to override this abstract method with another abstract method. You might need a `class FormValidationAction(action, ABC)` though to mark `FormValidationAction` as abstract

rasa_sdk/forms.py
```diff
@@ -689,3 +689,76 @@ def _get_entity_type_of_slot_to_fill(
...
+    ) -> List[EventType]:
+        """Validate slots by calling a validation function for each slot.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            A dictionary of `rasa_sdk.events.Event` instances.
```
wochinge, line 720:
```suggestion
            `SlotSet` events for every validated slot.
```

rasa_sdk/forms.py
```diff
@@ -689,3 +689,76 @@ def _get_entity_type_of_slot_to_fill(
...
+    ) -> List[EventType]:
+        """Execute the side effects of this action.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            A dictionary of `rasa_sdk.events.Event` instances.
```
wochinge, line 761: This is a list of events. Imo you don't have to specify a docstring as the docstring is already specified in the parent class.

rasa_sdk/forms.py
```diff
@@ -689,3 +689,76 @@ def _get_entity_type_of_slot_to_fill(
...
+        """
+        slots: Dict[Text, Any] = tracker.form_slots_to_validate()
+
+        for slot_name, slot_value in slots.items():
+            function_name = f"validate_{slot_name}"
+            validate_func = getattr(self, function_name, lambda *x: False)
```
wochinge, line 726: optional: I think checking for a validation function as early as possible is a little bit clearer
```suggestion
            validate_func = getattr(self, function_name, None)
            if not validate_func:
                warnings.warn(...)
                continue
```

tests/test_forms.py
```diff
@@ -1493,3 +1494,121 @@ async def test_submit(form_class: Type[FormAction]):
...
+        self,
+        slot_value: Any,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Dict[Text, Any]:
+        if slot_value == "correct_value":
+            return {
+                "slot2": "validated_value",
+            }
```
wochinge, line 1525: should at least return an empty dict in case the slot isn't valid, correct?

wochinge: same for `validate_slot1`

tests/test_forms.py
```diff
@@ -1493,3 +1494,121 @@ async def test_submit(form_class: Type[FormAction]):
...
+
+    # tracker with active form
+    tracker = Tracker(
+        "default",
+        {},
+        {},
+        [SlotSet("slot1", "correct_value"), SlotSet("slot2", "incorrect_value")],
+        False,
+        None,
+        {"name": "some_form", "validate": True, "rejected": False},
```
wochinge, line 1551: `validate` is deprecated. Should be `"is_interrupted": False` instead

changelog/288.improvement.md
```diff
@@ -0,0 +1,22 @@
...
+        self,
+        slot_value: Any,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Dict[Text, Any]:
+        if slot_value == "correct_value":
+            return {
+                "slot1": "validated_value",
+            }
```
wochinge, line 21: `validate_slot1` needs to set the slot value to `None` in case it isn't valid.

rasa_sdk/forms.py
```diff
@@ -689,3 +689,76 @@ def _get_entity_type_of_slot_to_fill(
...
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> List[EventType]:
+        """Validate slots by calling a validation function for each slot.
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the state tracker for the current user.
```
wochinge, line 717: ?
```suggestion
            tracker: the conversation tracker for the current user.
```

tests/test_forms.py
```diff
@@ -1493,3 +1494,121 @@ async def test_submit(form_class: Type[FormAction]):
...
+        "action_listen",
+    )
+
+    dispatcher = CollectingDispatcher()
+    events = await form.run(dispatcher=dispatcher, tracker=tracker, domain=None)
+    # check that the form was activated and validation was performed
```
wochinge, line 1557: this comment doesn't make sense as the form was already activated (same for test functions)
```suggestion
```

tests/test_forms.py
```diff
@@ -1493,3 +1494,121 @@ async def test_submit(form_class: Type[FormAction]):
...
+            SlotSet("slot3", "some_value"),
+        ],
+        False,
+        None,
+        {"name": "some_form", "validate": True, "rejected": False},
+        "action_listen",
+    )
+
+    dispatcher = CollectingDispatcher()
+    with pytest.warns(None):
```
wochinge, line 1607: ?
```suggestion
    with pytest.warns(UserWarning):
```

wochinge (commented):
> I made it work the way it worked before (in the current implementation events get updated in slots dictionary) but I still have no idea if it's the right behavior or not because there is nothing about it in the description.


Looks great 🥇 Just a few comments, then we're ready to go. Where would you expect a description?

wochinge:
@alwx I thought about whether the `validation_function` should return a `List[EventType]` (e.g. `SlotSet` for the validated slots) instead of the dictionary which we are currently returning. I think it makes sense to do it as you did and keep the old implementation which returns the plain `Dict` as users were happy with that and then there is one less thing for them to re-learn / change.

alwx:
@wochinge done

tests/test_forms.py
```diff
@@ -1493,3 +1494,123 @@ async def test_submit(form_class: Type[FormAction]):
...
+        slot_value: Any,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Dict[Text, Any]:
+        if slot_value == "correct_value":
+            return {
+                "slot3": "validated_value",
+            }
+        # this function doesn't return anything when the slot value is incorrect
```
alwx, line 1544: done this to test the warning

rasa_sdk/forms.py
```diff
@@ -689,3 +690,61 @@ def _get_entity_type_of_slot_to_fill(
...
+                )
+            else:
+                validation_output = validate_func(
+                    slot_value, dispatcher, tracker, domain
+                )
+
+            if validation_output:
+                slots.update(validation_output)
+            else:
+                warnings.warn(
```
wochinge, line 738: 🚀 Nice to catch this case 👍

tests/test_forms.py
```diff
@@ -1493,3 +1494,123 @@ async def test_submit(form_class: Type[FormAction]):
...
+        {"name": "some_form", "is_interrupted": False, "rejected": False},
+        "action_listen",
+    )
+
+    dispatcher = CollectingDispatcher()
+    events = await form.run(dispatcher=dispatcher, tracker=tracker, domain=None)
+    assert events == [SlotSet("slot3", "validated_value")]
+
+
+async def test_form_validation_action_attribute_error():
```
wochinge, line 1590:
```suggestion
async def test_form_validation_without_validate_function():
```

rasa_sdk/forms.py
```diff
@@ -689,3 +690,61 @@ def _get_entity_type_of_slot_to_fill(
...
+
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the conversation tracker for the current user.
+            domain: the bot's domain.
+        Returns:
+            `SlotSet` events for every validated slot.
+        """
+        slots: Dict[Text, Any] = tracker.form_slots_to_validate()
```
wochinge, line 714: I renamed it this morning (tagged you there)
```suggestion
        slots: Dict[Text, Any] = tracker.slots_to_validate()
```

wochinge (approved):
💯

PR #301, opened 2020-10-13
Redirections Redux

lunelson:
This PR changes the build path of the `docusaurus build` command to match the subpath configured as `baseUrl` in `docusarus.config.js`. Doing this avoids the additional rewrite of sub-directory paths back to the root, and enables content 301 redirections to be maintained in this repo, rather than in `rasa-website`.

This corresponds to [this PR on rasa-website](https://github.com/RasaHQ/rasa-website/pull/864), which should be merged first.

NOTE: the `/docs/static/_redirects` file has been replaced by a `/docs/netlify.toml` instead, because due to the changed build path it would now end up in a subfolder where it wouldn't work. Depending on how the build/deploy is scripted, `netlify.toml` may not work. In this case we could revert to `_redirects`, but would have to place the `_redirects` file manually at the root of the build, as part of the script, not in `/static/`.

### DIFF
--- a/docs/netlify.toml
+++ b/docs/netlify.toml
@@ -1,0 +1,5 @@
+# FIXME: this requires some feedback as to the 404, like in the website
+[[redirects]]
+  from = "/docs/action-server/*"
+  to = "/docs/action-server/index.html"
+  status = 404
--- a/docs/package.json
+++ b/docs/package.json
@@ -9,7 +9,7 @@
     "start": "yarn pre-build && yarn develop",
     "develop": "docusaurus start --port 3002",
     "pre-build": "yarn copy-md-files && yarn variables && yarn included-sources && yarn program-outputs",
-    "build": "docusaurus build",
+    "build": "docusaurus build --out-dir build/docs/action-server",
     "serve": "netlify dev --dir=build --port=5002",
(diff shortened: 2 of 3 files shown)


m-vdb:
@lunelson what do you mean by
> Depending on how the build/deploy is scripted, netlify.toml may not work

?
This is something that is deterministic and that we know right? Files are copied into the `documentation` branch and then Netlify uses the config that we saved in their UI to build the sites. What else do we need here?

lunelson:
>This is something that is deterministic and that we know right?

m-vdb:
@lunelson it's Netlify who takes care of the build. Github actions simply take care of updating the `documentation` branch. You can take a look at the config for Rasa for instance: https://app.netlify.com/sites/rasahq-docs-rasa-v2/settings/deploys

Do we need to put `netflify.toml` at the root of the repository in that case (instead of `docs/netlify.toml`)?

lunelson:
@m-vdb Great! No I think it's OK, because `Base directory: docs` is configured, this is where it will look for the `netlify.toml`

image

m-vdb:
perfect, and it's the same for rasa-x and rasa-sdk

m-vdb (approved):
⭐

Issue #161, opened 2020-03-27
Add --log-file option when using the action server


roniemartinez:
**Description of Problem**:
There is no option to save logs to a log file when using the actions server from the command line as oppose to API server with `--log-file` argument

```
% rasa run actions --help
usage: rasa run actions [-h] [-v] [-vv] [--quiet] [-p PORT]
                        [--cors [CORS [CORS ...]]] [--actions ACTIONS]
                        [--ssl-keyfile SSL_KEYFILE]
                        [--ssl-certificate SSL_CERTIFICATE]
                        [--ssl-password SSL_PASSWORD]

optional arguments:
  -h, --help            show this help message and exit
  -p PORT, --port PORT  port to run the server at (default: 5055)
  --cors [CORS [CORS ...]]
                        enable CORS for the passed origin. Use * to whitelist
                        all origins (default: None)
  --actions ACTIONS     name of action package to be loaded (default: None)
  --ssl-keyfile SSL_KEYFILE
                        Set the SSL certificate to create a TLS secured
                        server. (default: None)
  --ssl-certificate SSL_CERTIFICATE
                        Set the SSL certificate to create a TLS secured
                        server. (default: None)
  --ssl-password SSL_PASSWORD
                        If your ssl-keyfile is protected by a password, you
                        can specify it using this paramer. (default: None)

Python Logging Options:
  -v, --verbose         Be verbose. Sets logging level to INFO. (default:
                        None)
  -vv, --debug          Print lots of debugging statements. Sets logging level
                        to DEBUG. (default: None)
  --quiet               Be quiet! Sets logging level to WARNING. (default:
                        None)
```

**Overview of the Solution**:
Add option for the action server to save logs to a log file similar to what the API server can do.

**Definition of Done**:
- [ ] Can pass `--log-file` as command line parameter

sara-tagger:
Thanks for raising this issue, @erohmensing will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

erohmensing:
Hi @roniemartinez, i think this sounds like a promising enhancement. However one thing to think about is that when the SDK is running as an image, it runs via the `start` command https://github.com/RasaHQ/rasa-sdk/blob/master/Dockerfile#L40
so i guess that parameter would need to be available for both methods

roniemartinez:
[...]

PR #306, opened 2020-10-15
Logging to file for action server

dingusagar:
**Proposed changes**:
- Added command line argument option for the action server to save logs to a log file.


dingusagar:
Linked issue here #161

roniemartinez:
Awesome! Hope the project owners review this. 💯

sara-tagger:
Thanks for submitting a pull request 🚀 @melindaloubser1 will take a look at it as soon as possible ✨

iwt-kschoenrock:
Wouldn't it be better to use a [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler) directly? This would allow you to set `backupCount` to avoid eventually running out of space because of old logs.
Maybe let the user configure the number of backup files kept around?

dingusagar:
> Wouldn't it be better to use a [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler) directly? This would allow you to set `backupCount` to avoid eventually running out of space because of old logs.
> Maybe let the user configure the number of backup files kept around?

[RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler)  seems to be a better choice than [FileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.FileHandler). `backupCount` argument can be kept to 0 by default so that the default behaviour is a single file growing indefinitely. And user can specify `backupCount` as a command line argument. 

What do you think should be the value of `maxBytes` argument. Do you feel that needs to be configurable as well? @iwt-kschoenrock /  @roniemartinez 
Shall I add this enhancement to the PR. Any other concerns? @melindaloubser1

iwt-kschoenrock:
> [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler) seems to be a better choice than [FileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.FileHandler). `backupCount` argument can be kept to 0 by default so that the default behaviour is a single file growing indefinitely. And user can specify `backupCount` as a command line argument.
> 
> What do you think should be the value of `maxBytes` argument. Do you feel that needs to be configurable as well? @iwt-kschoenrock / @roniemartinez
> Shall I add this enhancement to the PR. Any concerns / thoughts? @melindaloubser1

I don't have any current examples of how big the Rasa Logs would be, maybe someone from Rasa could give an estimate? But I agree that `backuptCount` and `maxBytes` should be configurable.

This functionality is important for us, as we're looking for ways to configure ingestion of multi-line logs in Datadog, and the log file would probably be the easiest way to achieve it, apart from the nonplusultra solution, which would be allowing a custom log formatter or [log configuration file to be provided](https://docs.python.org/3/library/logging.config.html#configuration-file-format).

melindaloubser1:
The action server runs separately from the rasa server, so logging to files for custom actions should be configured on the action server itself (whether rasa-sdk or another action server). The changes look like they're configuring logging of all logs to a file (for the rasa server).

dingusagar:
> The action server runs separately from the rasa server, so logging to files for custom actions should be configured on the action server itself (whether rasa-sdk or another action server). The changes look like they're configuring logging of all logs to a file (for the rasa server). Can you clarify the desired behaviour of the change?

valid point. I have updated the code to configure logging only for action server. I gave "rasa_sdk" as the name for the named logger. so now the configuration is only for loggers within the moduel rasa_sdk. 

Could you review my latest changes? @melindaloubser1 
Thanks

melindaloubser1 (changes requested):
The logging to file works great, but it seems to set the logging level to `debug` regardless of what is passed as the log level - could you look into that?

dingusagar:
> The logging to file works great, but it seems to set the logging level to `debug` regardless of what is passed as the log level - could you look into that?

Ok I will look into it

dingusagar:
> The logging to file works great, but it seems to set the logging level to `debug` regardless of what is passed as the log level - could you look into that?

Hi @melindaloubser1 
I have pushed the fix for setting the log level of root logger of the project  from the command line args.

I noticed that the logger for matplotlib is set to WARN and logger for loglevel for sanic logger is set by environment variables. I have't changed any of those. But let me know if these loglevels also needs to controlled by the arguement passed in the command line args

wochinge:
fixing the `code_quality` step [here](https://github.com/RasaHQ/rasa-sdk/pull/354). Sorry for the inconvenience!

melindaloubser1 (changes requested):
Can you run  `make formatter` and apply the changes? The linting step is failing.

dingusagar:
> Can you run `make formatter` and apply the changes? The linting step is failing.

sure @melindaloubser1 
have made the changes for the errors given by `make lint`

melindaloubser1 (approved):

melindaloubser1:
Sorry one more thing, can you add a changelog entry please? You'll find instructions for that in the changelog/README.md

dingusagar:
> Sorry one more thing, can you add a changelog entry please? You'll find instructions for that in the changelog/README.md

added a changelog entry.

PR #329, opened 2020-11-04
Custom and dynamic slots

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7156
- users can now override a method `required_slots` to specify slots which should be filled by the form. The `FormValidationAction` will look for a method `extract_<slot_name>` to extract these slots.
- in case the users overrode `required_slots`, the `FormValidationAction` will automatically set the `requested_slot` to the next `required_slot` which is currently not filled. Users can override this behavior by customizing `request_next_slot`. By default `request_next_slot` also considers slots which were mapped in the domain (if there were any mapped ones). 
- introduce a utility function `utils.call_potential_coroutine` to get rid of our checks `if is coroutine ... else ...`

I will update the Rasa Open Source documentation in a separate Rasa Open Source PR.

### DIFF
--- a/changelog/329.feature.md
+++ b/changelog/329.feature.md
@@ -1,5 +1,5 @@
 Extend the `FormValidation` action with support for extracting slots. If you
-want to extract a custom slot, add the slots name to the list of `missing_slots`
+want to extract a custom slot, add the slots name to the list of `custom_slots`
 and add a method `extract<slot name>` to your action:
 
 ```python
@@ -13,14 +13,13 @@ class FormWithSlotExtractions(FormValidationAction):
     def name(self) -> Text:
         return "some_form"
 
-    async def missing_slots(
+    async def custom_slots(
         self,
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        all_required_slots = ["my slot"]
-        return [slot for slot in all_required_slots if not tracker.slots.get(slot)]
+        return ["my slot"]
 
     async def extract_my_slot(
         self,
@@ -31,5 +30,6 @@ class FormWithSlotExtractions(FormValidationAction):
         return {"my_slot": "some value"}
 ```
 
-If all `missing_slots` are filled, the action will automatically return an event
-to disable the form.
+If all `custom_slots` are filled, the action will automatically return an event
+to disable the form. If not every custom slot is filled, the SDK will return an event
+to Rasa Open Source to fill the first missing slot next.
--- a/changelog/329.feature.md
+++ b/changelog/329.feature.md
@@ -1,5 +1,5 @@
 Extend the `FormValidation` action with support for extracting slots. If you
-want to extract a custom slot, add the slots name to the list of `custom_slots`
+want to extract a custom slot, add the slots name to the list of `required_slots`
 and add a method `extract<slot name>` to your action:
 
 ```python
@@ -13,7 +13,7 @@ class FormWithSlotExtractions(FormValidationAction):
     def name(self) -> Text:
         return "some_form"
 
-    async def custom_slots(
+    async def required_slots(
         self,
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
@@ -30,6 +30,7 @@ class FormWithSlotExtractions(FormValidationAction):
         return {"my_slot": "some value"}
 ```
 
-If all `custom_slots` are filled, the action will automatically return an event
-to disable the form. If not every custom slot is filled, the SDK will return an event
+If all `required_slots` and all form slots which are specified in the domain are filled,
+the action will automatically return an event to disable the form. 
+If not every custom slot is filled, the SDK will return an event
 to Rasa Open Source to fill the first missing slot next.
--- a/changelog/329.feature.md
+++ b/changelog/329.feature.md
@@ -1,0 +1,35 @@
+Extend the `FormValidation` action with support for extracting slots. If you
+want to extract a custom slot, add the slots name to the list of `missing_slots`
+and add a method `extract<slot name>` to your action:
+
+```python
+from typing import Text, Dict, Any, List, Optional
+
+from rasa_sdk.forms import (
+    FormValidationAction,
+)
+
+class FormWithSlotExtractions(FormValidationAction):
+    def name(self) -> Text:
+        return "some_form"
+
+    async def missing_slots(
+        self,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Optional[List[Text]]:
+        all_required_slots = ["my slot"]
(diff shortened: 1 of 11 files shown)


wochinge:
@samsucik I imagined something along these lines. What do you think?

samsucik:
Looks good to me at a high level :-)

Would `next_requested_slot()` (in `FormValidationAction`) call `required_slots()` and attempt to select the next requested slot from the list that `required_slots()` returns?

wochinge:
> Would next_requested_slot() (in FormValidationAction) call required_slots() and attempt to select the next requested slot from the list that required_slots() returns?

I thought about it and decided that this would be something they should implement themselves. I want to avoid to change something for users who are just using this to validate their slots.

samsucik:
That's fair. Though I think we should _in some way_ communicate that the intended pattern (when one wants to change the required slots dynamically) is to call `next_requested_slot()` which then calls `required_slots()` where the logic for changing the set of required slots should be placed. What do you think about communicating this in the form of explanatory comments in the docstring of `next_requested_slot()` & `required_slots()` and possibly also some simple (commented out) code example in `next_requested_slot()`?

samsucik:
Also, I think that with `FormValidationAction` it gets confusing that if your `run()` doesn't return any requested slots (and by the same logic if your `next_requested_slot()` & `required_slots()` don't return any slots to be filled), **then your form will still continue** if it's got some unfilled slots from the set defined in the domain. From looking at others' code but also from my own experience, I think that it's easy to fall in the trap of seeing your subclass of `FormValidationAction` as the only class responsible for the given form, without realising that there's one more layer above this class and that your subclass of `FormValidationAction` is just a way to interrupt/alter the behaviour provided by that above layer.

wochinge:
Makes sense 👍  We can fill `required_slots` with the slot mappings from the domain by default and then always call `request_next_slot` on the next missing slot by default.

samsucik:
Hmmm, this is an interesting idea. While it hides the "meta layer" even more, I think it also makes it very clear to the user how `required_slots()` and `request_next_slot()` can be used and how exactly they fit into the whole pattern. One risk I can see is that if the user doesn't explicitly set `requested_slot` to any value in `run()`, they might be very surprised to see that some next slot gets requested anyway (in case all the slots -- as defined in the domain for the form -- aren't filled).

wochinge:
How about we only return a `requested_slot` **if** `required_slots` was overridden?

samsucik:
Sounds good to me! Together with an explanatory comment in `run()` (or also elsewhere) this should be clear enough I think.

wochinge:
@samsucik Could you please have another look? 

I split it up in `custom` slots and `missing_slots`. This way it's possible to implement a static version of `custom_slots` and still being able to request the next slot accordingly. It also allows for more precise overriding.

changelog/329.feature.md
```diff
@@ -0,0 +1,36 @@
...
+    def name(self) -> Text:
+        return "some_form"
+
+    async def required_slots(
+        self,
+        dispatcher: "CollectingDispatcher",
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Optional[List[Text]]:
+        return ["my slot"]
```
akelad, line 22: in which situations would you need to fill this out? you can specify slots without a slot mapping in the domain right?

akelad: ok i just realised why, in case you want to dynamically change the list of requested slots :D

akelad:
@wochinge anything in particular you want me to review? the example in the changelog looks good to me

changelog/329.feature.md
wochinge: you'd override for two use cases:
* dynamic slots
* custom slot mappings

wochinge:
@akelad I mainly wanted your feedback on the overall approach and as you requested this feature multiple times.

changelog/329.feature.md
akelad: ok so the custom slot mapping part i'm confused about: if you specify your slots without a slot mapping in the domain file, shouldn't it just use the `extract_` method for that slot?

akelad:
looks great to me, aside from the discussion we're having in one of the comments now :)

changelog/329.feature.md
wochinge: You mean that we would just try to extract every slot for which we have an `extract_` function defined? I think that would work fine if the slot mappings are static. It would become tricky in case your `required_slots` change dynamically

akelad: but would you list the "optional" required slots in the domain in the first place? if you didn't, i guess you wouldn't have the issue with the dynamically changing ones right?

wochinge: > but would you list the "optional" required slots in the domain in the first place? 

Nope, I wouldn't.

>  i guess you wouldn't have the issue with the dynamically changing ones right?

@samsucik interacted with users who had dynamically changing custom slot mappings. So that problem would still persist.

akelad: ok i'm super confused now - the way to extract the slot changes? how would that even work?

samsucik: I don't think "my" users wanted to dynamically change how slots are extracted. They wanted to dynamically change the list of required slots, but extraction was either given right in the domain or implemented using `extract_{slot}` (in either case, the implementation didn't change). The way I understand it is that this new implementation should use `extract_{slot}` whenever it exists for the given slot, and use the "default" extraction approach otherwise.

wochinge: Sounds good for you @akelad ?

akelad: yes, that makes sense for me! So just to summarise, for the requested slot method that means you only need to fill it out if you want to dynamically change requested slots, correct?

wochinge: Exactly. This includes the case that you want to ignore / override slot mappings defined in the domain (ideally they just leave the slot mappings empty in the domain)

rasa_sdk/utils.py
```diff
@@ -189,3 +186,12 @@ def update_sanic_log_level() -> None:
     logger.propagate = False
     error_logger.propagate = False
     access_logger.propagate = False
+
+
+async def call_potential_coroutine(
```
alwx, line 191: Nice approach, thanks for getting rid of all the annoying `if`s with this one!

rasa_sdk/forms.py
```diff
@@ -695,7 +678,104 @@ def _get_entity_type_of_slot_to_fill(
...
+        Args:
+            dispatcher: the dispatcher which is used to
+                send messages back to the user. Use
+                `dispatcher.utter_message()` for sending messages.
+            tracker: the state tracker for the current
+                user. You can access slot values using
+                `tracker.get_slot(slot_name)`, the most recent user message
+                is `tracker.latest_message.text` and any other
+                `rasa_sdk.Tracker` property.
+            domain: the bot's domain
```
alwx, line 722:
```suggestion
            domain: the bot's domain.
```

alwx (approved):
Looks good!

PR #338, opened 2020-11-13
slot extractions v2

wochinge:
**Proposed changes**:
- follow up of https://github.com/RasaHQ/rasa/pull/7233 and https://github.com/RasaHQ/rasa-sdk/pull/329
- `required_slots` receive the slots which were mapped in the domain as parameter to make it explicit for users that this is something which `required_slots` should return
- warn if a slot was mapped in the domain and has an additional custom mapping within the SDK.

### DIFF
--- a/changelog/329.feature.md
+++ b/changelog/329.feature.md
@@ -20,7 +20,7 @@ class FormWithSlotExtractions(FormValidationAction):
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        return slots_mapped_in_domain + ["my slot"]
+        return slots_mapped_in_domain + ["my_slot"]
 
     async def extract_my_slot(
         self,
--- a/changelog/329.feature.md
+++ b/changelog/329.feature.md
@@ -1,6 +1,6 @@
-Extend the `FormValidation` action with support for extracting slots. If you
-want to extract a custom slot, add the slots name to the list of `required_slots`
-and add a method `extract<slot name>` to your action:
+Extend `FormValidationAction` with support for extracting slots. If you
+want to extract an additional slot, add the slot's name to the list of `required_slots`
+and add a method `extract_<slot name>` to your action:
 
 ```python
 from typing import Text, Dict, Any, List, Optional
@@ -15,11 +15,12 @@ class FormWithSlotExtractions(FormValidationAction):
 
     async def required_slots(
         self,
+        slots_mapped_in_domain: List[Text],
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        return ["my slot"]
+        return slots_mapped_in_domain + ["my slot"]
 
     async def extract_my_slot(
         self,
@@ -30,7 +31,7 @@ class FormWithSlotExtractions(FormValidationAction):
         return {"my_slot": "some value"}
 ```
 
-If all `required_slots` and all form slots which are specified in the domain are filled,
+If all slots returned by `required_slots` are filled,
 the action will automatically return an event to disable the form. 
-If not every custom slot is filled, the SDK will return an event
+If not all required slots are filled, the SDK will return an event
 to Rasa Open Source to fill the first missing slot next.
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -657,7 +657,8 @@ class FormAction(Action):
         return f"FormAction('{self.name()}')"
 
     def _get_entity_type_of_slot_to_fill(
-        self, slot_to_fill: Optional[Text],
+        self,
+        slot_to_fill: Optional[Text],
     ) -> Optional[Text]:
         if not slot_to_fill:
             return None
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -744,20 +744,20 @@ class FormValidationAction(Action, ABC):
         domain: "DomainDict",
     ) -> List[EventType]:
         method_name = f"extract_{slot_name.replace('-', '_')}"
-        also_mapped_in_domain = slot_name in self.slots_mapped_in_domain(domain)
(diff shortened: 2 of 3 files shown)


rasa_sdk/forms.py
```diff
@@ -726,57 +726,90 @@ async def extract_custom_slots(
...
+        )
 
-        return custom_slots
+        if extracted:
+            return [SlotSet(slot, value) for slot, value in extracted.items()]
+
+        warnings.warn(
+            f"Cannot extract `{slot_name}`: make sure the extract method "
+            f"returns the correct output."
+        )
```
samsucik, line 777: Hmmm, I think this warning shouldn't be here. For any user utterance, we try to extract all slots from it. So it's natural that in most cases the call to an extractor won't actually extract anything, right?

rasa_sdk/forms.py
```diff
@@ -726,57 +726,90 @@ async def extract_custom_slots(
...
+        return []
 
     async def required_slots(
         self,
+        slots_mapped_in_domain: List[Text],
```
samsucik, line 782: @wochinge what do you think about renaming this to `slots_with_predefined_mappings` or similar? I'm not 100% certain and would like to get someone else's opinion, but it's true that in the docs I've been trying to remove the mentions of slots "mapped in the domain".

rasa_sdk/forms.py
```diff
@@ -726,57 +726,90 @@ async def extract_custom_slots(
...
             dispatcher: the dispatcher which is used to
                 send messages back to the user.
             tracker: the conversation tracker for the current user.
             domain: the bot's domain.
 
         Returns:
-            `None` in case this form doesn't extract any custom slots. Otherwise
-            returns the names of all custom slots which need to be filled.
+            Slot names which should be filled by the domain. By default it
+            returns the slot names which were mapped in the domain for this form.
```
samsucik, line 799:
```suggestion
            Slot names which should be filled by the form. By default it
            returns the slot names which are listed for this form in the domain 
            and use predefined mappings.
```

tests/test_forms.py
```diff
@@ -1679,6 +1707,7 @@ def name(self) -> Text:
 
         async def required_slots(
             self,
+            slot_mapped_in_domain: List[Text],
```
samsucik, line 1710:
```suggestion
            slots_mapped_in_domain: List[Text],
```

tests/test_forms.py
```diff
@@ -1736,6 +1765,7 @@ def name(self) -> Text:
...
```
samsucik, line 1768:
```suggestion
            slots_mapped_in_domain: List[Text],
```

tests/test_forms.py
```diff
@@ -1773,25 +1803,82 @@ async def extract_my_slot(
...
+    custom_slot = "my_slot"
+    unvalidated_value = "some value"
+
+    class TestFormValidationWithCustomSlots(FormValidationAction):
+        def name(self) -> Text:
+            return "some_form"
+
+        async def required_slots(
+            self,
+            slot_mapped_in_domain: List[Text],
```
samsucik, line 1827:
```suggestion
            slots_mapped_in_domain: List[Text],
```

samsucik (changes requested):
Looks good to me, I requested just tiny changes. But I think someone else should also have a look at the code, especially at the tests, because I might've missed something.

changelog/329.feature.md
```diff
@@ -15,11 +15,12 @@ class FormWithSlotExtractions(FormValidationAction):
 
     async def required_slots(
         self,
+        slots_mapped_in_domain: List[Text],
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        return ["my slot"]
+        return slots_mapped_in_domain + ["my slot"]
```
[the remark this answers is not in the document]
wochinge: very good catch! thanks!

rasa_sdk/forms.py
wochinge: I think in the documentation it made sense in combination with the link. I think in this case `domain` makes it clearer for now.

rasa_sdk/forms.py
wochinge: They need to return a dict of the form `{"slotname": <value>}` so the warning makes sense. Will adapt it a bit though.

samsucik: Okay, thanks for correcting me!

samsucik (approved):
Looks good to me 👍

rasa_sdk/forms.py
```diff
@@ -726,57 +726,93 @@ async def extract_custom_slots(
...
+        extract_method = getattr(self, method_name, None)
+
+        if not extract_method and not also_mapped_in_domain:
+            warnings.warn(
+                f"No method '{method_name}' found for slot "
+                f"'{slot_name}'. Skipping extraction for this slot."
+            )
+            return []
+
+        if not extract_method:
```
joejuzl, line 757: For these conditionals I find it hard to determine which combination of `also_mapped_in_domain` and `extract_method` lead to each branch as it requires keeping each previous `if` in your head. I think it would be clearer if you used either a traditional nested `if` structure, or re-stated both variables in each flat `if` statement.

rasa_sdk/forms.py
```diff
@@ -844,57 +880,20 @@ async def next_requested_slot(
...
-
-        all_required_slots = required_slots + [
-            domain_slot
-            for domain_slot in slots_mapped_in_domain
-            if domain_slot not in required_slots
-        ]
+        required_slots = await self.required_slots(
+            self.slots_mapped_in_domain(domain), dispatcher, tracker, domain
+        )
+        if required_slots and required_slots == self.slots_mapped_in_domain(domain):
```
joejuzl, line 886: count ordering be an issue here? Maybe a `set` is safer.

samsucik: Ordering matters here. The user might want to make a single change to the form: changing the order in which the slots are requested.

PR #339, opened 2020-11-13
7078/intent_from_fallback_helper

joejuzl:
#7078 Add a helper method to rasa-sdk for getting the next most likely user intent in case of nlu fallback.

### DIFF
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -272,28 +272,34 @@ class Tracker:
             self.slots[event["name"]] = event["value"]
             self.events.append(event)
 
-    def get_intent_if_not_fallback(self) -> Optional[Text]:
-        """Retrieves the intent the last user message would have been classified as
-        if it was not nlu_fallback
+    def get_intent_of_latest_message(
+        self, skip_fallback_intent: bool = True
+    ) -> Optional[Text]:
+        """Retrieves the intent the last user message.
+
+        Args:
+            skip_fallback_intent: Optionally skip the nlu_fallback intent
+                and return the next.
         """
-        last_user_action = self.get_last_event_for("user")
-        if not last_user_action:
-            warnings.warn("No last user action.")
+        latest_message = self.latest_message
+        if not latest_message:
             return None
 
-        parse_data = last_user_action["parse_data"]
-        intent = parse_data["intent"]
-        if intent["name"] != NLU_FALLBACK_INTENT_NAME:
-            warnings.warn("Last user action was not nlu fallback")
+        intent_ranking = latest_message.get("intent_ranking")
+        if not intent_ranking:
             return None
 
-        intent_ranking = parse_data["intent_ranking"]
-        if len(intent_ranking) < 2:
-            warnings.warn("No other potential intents.")
-            return None
-
-        second_highest_intent = intent_ranking[1]
-        return second_highest_intent["name"]
+        highest_randing_intent = intent_ranking[0]
+        if (
+            highest_randing_intent["name"] == NLU_FALLBACK_INTENT_NAME
+            and skip_fallback_intent
+        ):
+            if len(intent_ranking) >= 2:
+                return intent_ranking[1]["name"]
+            else:
+                return None
+        else:
+            return highest_randing_intent["name"]
 
 
 class Action:
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -13,6 +13,7 @@ if typing.TYPE_CHECKING:  # pragma: no cover
 logger = logging.getLogger(__name__)
 
 ACTION_LISTEN_NAME = "action_listen"
+NLU_FALLBACK_INTENT_NAME = "nlu_fallback"
 
 
 class Tracker:
@@ -270,6 +271,29 @@ class Tracker:
                 continue
             self.slots[event["name"]] = event["value"]
             self.events.append(event)
+
+    def get_intent_if_not_fallback(self) -> Optional[Text]:
+        """Retrieves the intent the last user message would have been classified as
+        if it was not nlu_fallback
+        """
+        last_user_action = self.get_last_event_for("user")
+        if not last_user_action:
+            warnings.warn("No last user action.")
+            return None
+
+        parse_data = last_user_action["parse_data"]
+        intent = parse_data["intent"]
+        if intent["name"] != NLU_FALLBACK_INTENT_NAME:
+            warnings.warn("Last user action was not nlu fallback")
+            return None
+
+        intent_ranking = parse_data["intent_ranking"]
+        if len(intent_ranking) < 2:
+            warnings.warn("No other potential intents.")
+            return None
+
+        second_highest_intent = intent_ranking[1]
+        return second_highest_intent["name"]
 
 
 class Action:
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -289,9 +289,9 @@ class Tracker:
         if not intent_ranking:
             return None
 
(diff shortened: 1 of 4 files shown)


wochinge:
Thanks for the quick PR! Will have a look on Monday 🚀

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,29 @@ def add_slots(self, slots: List[EventType]) -> None:
             self.slots[event["name"]] = event["value"]
             self.events.append(event)
 
+    def get_intent_if_not_fallback(self) -> Optional[Text]:
+        """Retrieves the intent the last user message would have been classified as
+        if it was not nlu_fallback
+        """
+        last_user_action = self.get_last_event_for("user")
```
wochinge, line 279: optional: You can also use `tracker.latest_message` which seems a bit more straightforward.

wochinge, line 279: I think `action` might be confusing within the Rasa Domain, as `action`s are usually something the assistant does.
```suggestion
        last_user_message = self.get_last_event_for("user")
```

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,29 @@ def add_slots(self, slots: List[EventType]) -> None:
...
+        if not last_user_action:
+            warnings.warn("No last user action.")
+            return None
+
+        parse_data = last_user_action["parse_data"]
```
wochinge, line 284: There should always be a `parse_data`, but I think this would be more robust (who knows what the API gets fed)
```suggestion
        parse_data = last_user_action.get("parse_data", {})
```

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,29 @@ def add_slots(self, slots: List[EventType]) -> None:
             self.slots[event["name"]] = event["value"]
             self.events.append(event)
 
+    def get_intent_if_not_fallback(self) -> Optional[Text]:
```
wochinge, line 275: I think there is no need to issue `warnings` as users don't know beforehand which intent they are expecting. Having no `nlu_fallback` intent present is hence neither unexpected state nor wrong usage. What do you think?

wochinge, line 275: How about something like this?
```suggestion
    def get_intent_of_last_message(self, skip_fallback_intent: bool = True) -> Optional[Text]:
```

tests/test_tracker.py
```diff
@@ -133,3 +132,41 @@ def test_get_extracted_slots_with_no_active_loop():
...
+            UserUttered(
+                text="Some Fallback",
+                parse_data={
+                    "intent": {"name": "nlu_fallback", "confidence": 0.9},
+                    "intent_ranking": [{"name": "nlu_fallback", "confidence": 0.9}],
+                },
+            )
+        ]
+    )
+    assert not tracker.get_intent_if_not_fallback()
```
wochinge, line 155: optional: I'm personally favoring to test one use case per test function. Hence, I'd either use `pytest.mark.parametrize` for the first two `asserts` and a separate test function for the third `assert` or using `pytest.mark.parametrize` for all three and add a comment about the use case to each input (e.g. `test with empty tracker`, `no fallback after last message`, `last user message triggered fallback` )

wochinge (commented):
Thanks for jumping on this! 💯

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,35 @@ def add_slots(self, slots: List[EventType]) -> None:
...
+        if (
+            highest_randing_intent["name"] == NLU_FALLBACK_INTENT_NAME
+            and skip_fallback_intent
+        ):
+            if len(intent_ranking) >= 2:
+                return intent_ranking[1]["name"]
+            else:
+                return None
+        else:
+            return highest_randing_intent["name"]
```
wochinge, line 302:
```suggestion
        highest_ranking_intent = intent_ranking[0]
        if (
            highest_ranking_intent["name"] == NLU_FALLBACK_INTENT_NAME
            and skip_fallback_intent
        ):
            if len(intent_ranking) >= 2:
                return intent_ranking[1]["name"]
            else:
                return None
        else:
            return highest_ranking_intent["name"]
```

tests/test_tracker.py
```diff
@@ -133,3 +132,45 @@ def test_get_extracted_slots_with_no_active_loop():
         "some_other": "some_value2",
         "my_slot": "some_value",
     }
+
+
+def test_get_intent_of_latest_message_with_missing_data():
```
wochinge, line 137: should we use the `NLU_FALLBACK_INTENT_NAME` constant in the tests?

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,35 @@ def add_slots(self, slots: List[EventType]) -> None:
...
             self.events.append(event)
 
+    def get_intent_of_latest_message(
+        self, skip_fallback_intent: bool = True
+    ) -> Optional[Text]:
+        """Retrieves the intent the last user message.
+
+        Args:
+            skip_fallback_intent: Optionally skip the nlu_fallback intent
+                and return the next.
```
wochinge, line 282: We should also describe `Returns:`

rasa_sdk/interfaces.py
```diff
@@ -271,6 +272,35 @@ def add_slots(self, slots: List[EventType]) -> None:
...
+        """
+        latest_message = self.latest_message
+        if not latest_message:
+            return None
+
+        intent_ranking = latest_message.get("intent_ranking")
```
wochinge, line 288: optional: pure style preference, but we could also do something like

```python
        intent_ranking = latest_message.get("intent_ranking", [])

        return next(
            (
                intent["name"]
                for intent in intent_ranking
                if not skip_fallback_intent
                or intent["name"] != NLU_FALLBACK_INTENT_NAME
            ),
            None,
        )
```

wochinge (commented):
Left a few more comments, but nothing critical! 🚀 

@erohmensing Is it enough to document the new helper in the docs here or is it better within the Rasa Open Source documentation? Having separate docs for the SDK is a bit new to me 😆 

@joejuzl And we need a changelog please.

rasa_sdk/interfaces.py
joejuzl: 🤦

erohmensing:
yep, we should add it to the changelog here (functions the same with the automatic changelog), and document it on this page: https://github.com/RasaHQ/rasa-sdk/blob/master/docs/docs/sdk-tracker.mdx

rasa_sdk/interfaces.py
joejuzl: I would tend to avoid using an iterator/generator if there is only max 2 elements that will ever be accessed.

wochinge: 👍 because it's slower?

wochinge (approved):
🚀

rasa_sdk/interfaces.py
joejuzl: No, I don't think it is slower in any meaningful way, it's a generator so it will not iterate the whole list (as it would if it was list comprehension).

I just think it adds a cognitive burden as the reader tends to try and understand the whole potential generator output, even if they know it's only the first element then get's retrieved.

Plus it's just a rule of thumb I use, that if there's less than three things, don't use a loop or list.

Although this is all quite subjective - so happy to have my mind changed!

wochinge: Thanks for the detailed explanation! I just wanted to understanding it a bit more. No strong opinion on it from my side 👍

joejuzl: Always happy to discuss stuff like this :)

Issue #348, opened 2020-11-19
Add headers to the domain to support OpenTelemetry context injection


rgstephens:
**Description of Problem**:
OpenTelemetry tracing passes context information via http headers. By adding the http headers to the domain dictionary, action server code can trace a rasa transaction from Rasa OSS to the action server.

**Overview of the Solution**:
Attach the http headers to the domain dictionary.

**Definition of Done**:
- [ ] Tests are added
- [x] Feature is documented

rgstephens:
This issue should stay open. We continue to get requests for this.

PR #349, opened 2020-11-19
Add headers to domain

rgstephens:
**Proposed changes**:
- Add headers to domain as described in #348


changelog/348.improvement.rst
```diff
@@ -0,0 +1 @@
+Add the http headers to the existing domain object so that OpenTelemetry context information can be passed to the action server.
```
wochinge, line 1: * how is the header used? That would be interesting to understand for me as I think there could be other options to the necessary information.
* And do we need all the headers?

rgstephens: - The header is needed for the trace context info. The w3c standard description is [here](https://www.w3.org/TR/trace-context/#trace-context-http-headers-format).
- We could pass only that w3c trace context but would also be good to pass the [opentracing header](https://www.jaegertracing.io/docs/1.21/client-libraries/#key) since we're in a transition period.

wochinge: Thanks for the explanation!

rasa_sdk/endpoint.py
```diff
@@ -89,6 +89,7 @@ async def health(_) -> HTTPResponse:
     async def webhook(request: Request) -> HTTPResponse:
         """Webhook to retrieve action calls."""
         action_call = request.json
+        action_call["domain"]["headers"] = request.headers
```
wochinge, line 92: I think the ideal solution would be to pass something like `request_context` as extra parameter to `run`. However, this would require quite a couple of changes to avoid breaking things. 

In my opinion storing this information in the `tracker` would be the next best solution as from my perspective the `tracker` _tracks_ the conversation state and so it makes sense to include the rest of the tracking here. Another things: How about keep this on a branch and cutting an alpha with your changes? This would not affect the official code but you could still go forward?

wochinge: And a simple test would in my opinion be nice 🙌🏻

rgstephens: I'm not following how the `request_context` parm would work. Is the idea that tracing info would only be passed if this option was enabled?

Including it in the tracker would add a lot of noise. These structures could get big and are not related to the conversation events except for tracing purposes. The tracing data would then be added to both the tracing system, tracker_store and on to the message queue.

wochinge: > I'm not following how the request_context parm would work. Is the idea that tracing info would only be passed if this option was enabled?

I meant adding an additional parameter which stores the tracing data instead of adding it to the existing `domain` parameter.

> Including it in the tracker would add a lot of noise. 

This would only be in the `tracker` within the SDK, wouldn't it? Rasa Open Source wouldn't see it as part of the tracker. In addition, only the `tracker["events"]` property is saved to tracker stores / message queue etc.

melindaloubser1:
Closing as stale

PR #363, opened 2020-12-07
updated poetry to 1.1.4

tmbo:
**Proposed changes**:
- updated poetry to 1.1.4
- needed to issues with dataclasses (used in transformers lib in rasa) https://github.com/python-poetry/poetry/issues/2386
- update development dependencies
- added additional build step to update buildtools (setuptools & pip) - required by newer poetry versions


joejuzl:
This appears to be breaking when installing the dependencies for the tests...

tmbo:
this seems to be an issue with an old pip version being used, I've added a separate build step to update pip and seems to be working now: https://github.com/python-poetry/poetry/issues/3329#issuecomment-724081350 

also increased dependency ranges (updates from depandabot) since I needed to regenerate the lockfile anyways

tmbo:
already everything working now - ready for another review @joejuzl

.github/workflows/continous-integration.yml
```diff
@@ -46,6 +46,9 @@ jobs:
         key: ${{ runner.os }}-poetry-3.7-${{ hashFiles('**/poetry.lock') }}
         restore-keys: ${{ runner.os }}-poetry-
 
+    - name: Update Buildtools 🥄
```
joejuzl, line 49: I wonder if we want these in the other project repos too?

joejuzl (approved):

tmbo:
🛑  can't be merged yet, build system requires the poetry version which isn't mentioned anywhere after switching the builder...https://github.com/python-poetry/poetry/issues/3469

wochinge:
Can we please wrap this up? Will release 2.3 with the old version.

tmbo:
Well, we need to solve the problem of the missing poetry version. we need to find a good spot to store that, until poetry hopefully comes up with their way of handling this. not sure if we can just add a random property to e.g. the pyproject file

wochinge:
Why aren't we running in the same issues on Rasa Open Source / Rasa X?

tmbo:
taking a look at the code it seems like Joe fixed it for Rasa Open Source by writing the poetry version into a separate file. https://github.com/RasaHQ/rasa/pull/7502/files so that code should just be ported to the sdk

joejuzl:
I will port the changes.

tmbo:
you'r the best 🚀 thanks a lot

pyproject.toml
```diff
@@ -67,23 +67,23 @@ python = ">=3.6,<3.9"
...
 towncrier = "^19.2.0"
-toml = "^0.10.0"
+toml = "^0.10.2"
 pep440-version-utils = "^0.3.0"
 semantic_version = "^2.8.5"
 typing-extensions = "^3.7.4"
-mypy = "^0.782"
+mypy = "^0.790"
 
 # uvloop >= 0.15.0 is not compatible with Python 3.6 and not available on Windows
```
wochinge, line 88: is `poetry>1.10` smart enough to detect the incompatibility of `uvloop==0.15.0` with Python 3.6 itself?

pyproject.toml
```diff
@@ -67,23 +67,23 @@ python = ">=3.6,<3.9"
...
-coveralls = "^2.1.2"
-pytest = "^5.4.1"
+coveralls = "^2.2.0"
+pytest = "^6.1.2"
 black = "^20.8b1"
 flake8 = "^3.8.2"
 flake8-docstrings = "^1.5.0"
 pytest-sanic = "^1.6.1"
-questionary = "^1.8.1"
+questionary = ">=1.7.0,<1.9.0"
```
wochinge, line 80: this is incompatible with what's specified in `rasa`. And why is this a dev dependency in the first place?

pyproject.toml
```diff
@@ -67,23 +67,23 @@ python = ">=3.6,<3.9"
...
 sanic-cors = "^0.10.0"
-requests = "^2.23"
+requests = ">=2.23.0,<2.26.0"
 
 [tool.poetry.dev-dependencies]
 pytest-cov = "^2.10.1"
-coveralls = "^2.1.2"
-pytest = "^5.4.1"
+coveralls = "^2.2.0"
+pytest = "^6.1.2"
```
wochinge, line 75: that's also different from Rasa Open Source which is fine, but I think for ease of development it would be great if these are similar

wochinge (commented):
Thanks for wrapping this up 🙌🏻

pyproject.toml
joejuzl: it blows up if only `markers = "sys_platform != 'win32'"` is there...

pyproject.toml
joejuzl: it's used in the release script.

wochinge (approved):
💯

PR #426, opened 2021-03-31
Updated code related to forms with the `required_slots` keyword

Imod7:
**Proposed changes**:
This PR is related to the changes implemented [here](https://github.com/RasaHQ/rasa/pull/7931) where we added the keyword `required_slots` in forms and we also deprecated the lack of it.

- Added the `required_slots` keyword where the form slots are returned.
- Updated the tests that have a domain with a form example so that it includes the keyword `required_slots` after the form name.

- [x] updated the tests that use a domain with a form
- [x] added a deprecation changelog (please check [changelog](https://github.com/RasaHQ/rasa-sdk/tree/main/changelog) for instructions)

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -19,7 +19,3 @@ jobs:
         # typo_ci_license_key: ${{ secrets.TYPO_CI_LICENSE_KEY }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-# Any files/folders we should ignore?
-excluded_files:
-  - "rasa_sdk/**/*"
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -19,3 +19,7 @@ jobs:
         # typo_ci_license_key: ${{ secrets.TYPO_CI_LICENSE_KEY }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+# Any files/folders we should ignore?
+excluded_files:
+  - "rasa_sdk/**/*"
--- a/changelog/261.improvement.md
+++ b/changelog/261.improvement.md
@@ -1,0 +1,2 @@
+Under each form of the domain, the definition of slot mappings without the keyword `required_slots` is now deprecated. Please always use the 
+keyword `required_slots` right before the slot mappings definition of each form in the `domain.yml` file. 
--- a/changelog/261.improvement.md
+++ b/changelog/261.improvement.md
@@ -1,2 +1,1 @@
-Under each form of the domain, the definition of slot mappings without the keyword `required_slots` is now deprecated. Please always use the 
-keyword `required_slots` right before the slot mappings definition of each form in the `domain.yml` file. 
+Adding support for forms with the `required_slots` key which is introduced as part of Rasa Open Source 2.6.
--- a/changelog/261.removal.md
+++ b/changelog/261.removal.md
@@ -1,2 +1,0 @@
-Under each form of the domain, the definition of slot mappings without the keyword ``required_slots`` is now deprecated. Please always use the 
-keyword ``required_slots`` right before the slot mappings definition of each form in the ``domain.yml`` file. 
--- a/changelog/261.removal.md
+++ b/changelog/261.removal.md
@@ -1,0 +1,2 @@
+Under each form of the domain, the definition of slot mappings without the keyword ``required_slots`` is now deprecated. Please always use the 
+keyword ``required_slots`` right before the slot mappings definition of each form in the ``domain.yml`` file. 
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -818,6 +818,11 @@ class FormValidationAction(Action, ABC):
             Slot names mapped in the domain.
         """
         form = domain.get("forms", {}).get(self.form_name(), {})
+        # In forms now, the keyword `required_slots` should always precede
+        # the definition of slot mappings and the lack of it is deprecated.
+        # So, we add the following check here to make sure we support
+        # the old (without keyword) and new (with keyword) format
+        # of the slot mappings definition.
         if "required_slots" in form:
             required_slots = form.get("required_slots", {}).keys()
         else:
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -817,12 +817,12 @@ class FormValidationAction(Action, ABC):
         Returns:
             Slot names mapped in the domain.
         """
-        return list(
-            domain.get("forms", {})
-            .get(self.form_name(), {})
-            .get("required_slots", {})
-            .keys()
-        )
+        form = domain.get("forms", {}).get(self.form_name(), {})
+        if "required_slots" in form:
+            required_slots = form.get("required_slots", {}).keys()
+        else:
+            required_slots = form.keys()
+        return list(required_slots)
 
     async def validate(
         self,
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -817,7 +817,12 @@ class FormValidationAction(Action, ABC):
         Returns:
             Slot names mapped in the domain.
         """
-        return list(domain.get("forms", {}).get(self.form_name(), {}).keys())
+        return list(
+            domain.get("forms", {})
+            .get(self.form_name(), {})
+            .get("required_slots", {})
+            .keys()
+        )
 
     async def validate(
         self,
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -1573,7 +1573,9 @@ async def test_form_validation_action():
         events = await form.run(
             dispatcher=dispatcher,
             tracker=tracker,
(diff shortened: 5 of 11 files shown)


Imod7:
Should the deprecation changelog in rasa-sdk, [here](https://github.com/RasaHQ/rasa-sdk/pull/426/files), have the same wording as the corresponding [one](https://github.com/RasaHQ/rasa/pull/7931/files) I have in rasa-oss ? If yes, which one have a better wording ?

rasa_sdk/forms.py
```diff
@@ -817,7 +817,12 @@ def slots_mapped_in_domain(self, domain: "DomainDict") -> List[Text]:
         Returns:
             Slot names mapped in the domain.
         """
-        return list(domain.get("forms", {}).get(self.form_name(), {}).keys())
+        return list(
+            domain.get("forms", {})
+            .get(self.form_name(), {})
+            .get("required_slots", {})
```
wochinge, line 823: can we please still try to retrieve the slots from the the old format in case somebody uses the new SDK version with an old Rasa Open Source version? 🙌🏻

changelog/261.removal.md
```diff
@@ -0,0 +1,2 @@
+Under each form of the domain, the definition of slot mappings without the keyword ``required_slots`` is now deprecated. Please always use the 
+keyword ``required_slots`` right before the slot mappings definition of each form in the ``domain.yml`` file.
```
wochinge, line 2:
```suggestion
Under each form of the domain, the definition of slot mappings without the keyword `required_slots` is now deprecated. Please always use the 
keyword `required_slots` right before the slot mappings definition of each form in the `domain.yml` file. 
```

changelog/261.removal.md
```diff
@@ -0,0 +1,2 @@
+Under each form of the domain, the definition of slot mappings without the keyword ``required_slots`` is now deprecated. Please always use the
```
wochinge, line 1: I'd say this is an `improvement` as it makes the SDK compatible with the the new Rasa Open Source release. You're not deprecating anything here.

wochinge (commented):
💯

rasa_sdk/forms.py
Imod7: You are so right! 👌🏾 I thought that since we add it in the `domain` in rasa-oss that it will always be there but I did not think about the older versions of Rasa. 🤦🏽‍♀️
I adapted the code and I also tested it by removing (temporarily) the `required_slots` keyword from a test. So it works with / without the `required_slots` keyword now.

Imod7:
I do not understand the typo mentioned [here](https://github.com/RasaHQ/rasa-sdk/pull/426/files)

wochinge:
the typo appears to be a false positive. @melindaloubser1 Where can we define the exceptions for this?

rasa_sdk/forms.py
```diff
@@ -817,7 +817,12 @@ def slots_mapped_in_domain(self, domain: "DomainDict") -> List[Text]:
         Returns:
             Slot names mapped in the domain.
         """
-        return list(domain.get("forms", {}).get(self.form_name(), {}).keys())
+        form = domain.get("forms", {}).get(self.form_name(), {})
+        if "required_slots" in form:
```
wochinge, line 821: let's add a comment why we have to do this to the code to make it clearer for developers which read the code in the future

rasa_sdk/forms.py
```diff
@@ -817,7 +817,12 @@ def slots_mapped_in_domain(self, domain: "DomainDict") -> List[Text]:
...
+            required_slots = form.get("required_slots", {}).keys()
+        else:
+            required_slots = form.keys()
```
wochinge, line 824: can we please add a test for this behavior?

melindaloubser1:
It should be in a  .typo-ci.yml file in the root of the repo - not sure if it somewhere got deleted here, or was never configured? Could start by copying Rasa's: https://github.com/RasaHQ/rasa/blob/main/.typo-ci.yml

rasa_sdk/forms.py
Imod7: I just added a case in `pytest parametrize` for the function `test_ask_for_next_slot` so it now tests a form with the new format (with `required_slots` keyword) and the old format (without the keyword).

Imod7 (commented):
I could not find a `typo-ci `yaml file in rasa-sdk but I found a `spellcheck` yaml file [here](https://github.com/RasaHQ/rasa-sdk/blob/main/.github/workflows/spellcheck.yml) where I added an `excluded folder` as it is in `Rasa OSS / typo-ci `. Let's see if it works.

.github/workflows/spellcheck.yml
```diff
@@ -19,3 +19,7 @@ jobs:
         # typo_ci_license_key: ${{ secrets.TYPO_CI_LICENSE_KEY }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+# Any files/folders we should ignore?
+excluded_files:
+  - "rasa_sdk/**/*"
```
wochinge, line 25: Thanks for the change. Adding an empty line at the end (https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file).

```suggestion
  - "rasa_sdk/**/*"
```

.github/workflows/spellcheck.yml
```diff
@@ -19,3 +19,7 @@ jobs:
         # typo_ci_license_key: ${{ secrets.TYPO_CI_LICENSE_KEY }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+# Any files/folders we should ignore?
```
wochinge, line 23: maybe ignoring all `*.py` files would be neat as well (you can maybe take the `spellcheck.yml` from the Rasa Open Source repository as inspiration). This goes beyond the scope of this PR so feel free to just create an issue for this or do it in a separate PR

tests/test_forms.py
```diff
@@ -1916,10 +1928,17 @@ async def extract_my_slot(
...
             [SlotSet(REQUESTED_SLOT, None)],
         ),
         # `required_slots` was not overridden - Rasa Open Source will request next slot.
+        # slot mappings with the `required_slots` keyword preceding them (new format)
+        (
+            ["another_slot"],
+            {"forms": {"some_form": {"required_slots": {"another_slot": []}}}},
+            [],
+        ),
+        # slot mappings without the `required_slots` keyword preceding them (old format)
```
wochinge, line 1941: 🚀

rasa_sdk/forms.py
wochinge: Slick!

changelog/261.improvement.md
```diff
@@ -0,0 +1,2 @@
+Under each form of the domain, the definition of slot mappings without the keyword `required_slots` is now deprecated. Please always use the
```
wochinge, line 1: If I put myself in the shoe of a user I find this a tiny bit cryptic. How about something like: "Adding support for forms with the `required_slots` key which is introduced as part of Rasa Open Source 2.6."?

wochinge (approved):
Ready toooo merge 🚀 (please see the two comments I made)

.github/workflows/spellcheck.yml
Imod7: nice! I didn't know! thanks for sharing. In the same I found also this https://stackoverflow.com/questions/11597901/why-are-trailing-commas-allowed-in-a-list

.github/workflows/spellcheck.yml
Imod7: Maybe you mean the [typo-ci.yml](https://github.com/RasaHQ/rasa/blob/main/.typo-ci.yml) that Melinda mentioned previously and contains the `*.py` that you are saying? Because in Rasa-OSS there is : 
- [typo-ci.yml](https://github.com/RasaHQ/rasa/blob/main/.typo-ci.yml) which has these type of file checks and
- [spellcheck.yml](https://github.com/RasaHQ/rasa/blob/main/.github/workflows/spellcheck.yml) which does not contain file related checks. 

Now that I see this again I think I did something wrong because it seems there are different files for different things : 
- `spellcheck.yml` as it is described [here](https://github.com/marketplace/actions/github-spellcheck) is "_...for spell checking in GitHub Actions._"
- `typo-ci.yml` as it is described [here](https://typoci.com/documentation) is for "_...checks Pull Requests for spelling mistakes every time you push your code up to GitHub_"

and I copied a file check / rule (`"rasa_sdk/**/*"`) from `typo-ci.yml` (rasa-oss) to `spellcheck.yml` (rasa-sdk) because in rasa-sdk there is no typo-ci file (at least I could not find one). But maybe that is not correct and I should create a typo-ci.yml in the root folder of rasa-sdk and just copy all the checks/rules from rasa-oss as Melinda suggested.

wochinge: good catch!

wochinge: >  My only concern is why this file was not already there? Was it removed for a reason and I should not add it?

I guess nobody ran into this 🤷🏻

wochinge (approved):
Great catch with the spellcheck file!

PR #441, opened 2021-05-06
preserving the order of slots to be validated (rather than reversed) + making updated slot value immediately visible to extract/validate of subsequent slots.

cmanovit:
According to the following question in the forum, slots should be validated in the order they are defined in required_slots method.

https://forum.rasa.com/t/what-is-the-order-of-execution-of-validate-slot-name-functions-in-formaction-feature/14562

But what I observed is that the order is actually reversed from that specified in my required_slots in FormValidationAction. I traced it to this code and found that the order gets reversed here. 

[For the second commit] Also, when there are multiple slots to be validated in a single message, it is easier to understand and code if during validation of the second slot we see the updated value of the first slot.

Example: How is the weather in \[San Francisco\](city), \[California\](state)?

In this case, we may want to validate "state" before "city". Our database may use abbreviation for state name, so after validating the "state", we may want to convert it to "CA". And when we validate the "city", we may have to use it in conjunction with "CA" to look up from our database. Without the proposed change, the slot "state" would still have its original value "California" during validation of the slot "city".

**Proposed changes**:
- ...

### DIFF
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -864,6 +864,9 @@ class FormValidationAction(Action, ABC):
 
             if isinstance(validation_output, dict):
                 slots.update(validation_output)
+                # for sequential consistency, also update tracker
+                # to make changes visible to subsequent validate_{slot_name}
+                tracker.slots.update(validation_output)
             else:
                 warnings.warn(
                     f"Cannot validate `{slot_name}`: make sure the validation method "
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -730,16 +730,21 @@ class FormValidationAction(Action, ABC):
         Returns:
             `SlotSet` for any extracted slots.
         """
-        custom_slots = []
+        custom_slots = {}
         slots_to_extract = await self.required_slots(
             self.slots_mapped_in_domain(domain), dispatcher, tracker, domain
         )
 
         for slot_name in slots_to_extract:
-            custom_slots += await self._extract_slot(
+            extraction_output = await self._extract_slot(
                 slot_name, dispatcher, tracker, domain
             )
-        return custom_slots
+            custom_slots.update(extraction_output)
+            # for sequential consistency, also update tracker
+            # to make changes visible to subsequent extract_{slot_name}
+            tracker.slots.update(extraction_output)
+
+        return [SlotSet(slot, value) for slot, value in custom_slots.items()]
 
     async def _extract_slot(
         self,
@@ -747,7 +752,7 @@ class FormValidationAction(Action, ABC):
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
-    ) -> List[EventType]:
+    ) -> Dict[Text, Any]:
         method_name = f"extract_{slot_name.replace('-', '_')}"
         slot_mapped_in_domain = slot_name in self.slots_mapped_in_domain(domain)
         extract_method = getattr(self, method_name, None)
@@ -758,9 +763,7 @@ class FormValidationAction(Action, ABC):
                     f"No method '{method_name}' found for slot "
                     f"'{slot_name}'. Skipping extraction for this slot."
                 )
-                return []
-            else:
-                return []
+            return {}
 
         if extract_method and slot_mapped_in_domain:
             warnings.warn(
@@ -776,13 +779,13 @@ class FormValidationAction(Action, ABC):
         )
 
         if isinstance(extracted, dict):
-            return [SlotSet(slot, value) for slot, value in extracted.items()]
+            return extracted
 
         warnings.warn(
             f"Cannot extract `{slot_name}`: make sure the extract method "
             f"returns the correct output."
         )
-        return []
+        return {}
 
     async def required_slots(
         self,
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -259,7 +259,7 @@ class Tracker:
                 # checked all potential slot candidates.
                 break
 
-        for event in self.events[len(self.events)-count:]:
+        for event in self.events[len(self.events) - count :]:
             slots[event["name"]] = event["value"]
 
         return slots
--- a/rasa_sdk/interfaces.py
+++ b/rasa_sdk/interfaces.py
@@ -247,16 +247,20 @@ class Tracker:
         """
 
         slots: Dict[Text, Any] = {}
+        count: int = 0
 
         for event in reversed(self.events):
             # The `FormAction` in Rasa Open Source will append all slot candidates
             # at the end of the tracker events.
             if event["event"] == "slot":
-                slots[event["name"]] = event["value"]
+                count += 1
             else:
                 # Stop as soon as there is another event type as this means that we
                 # checked all potential slot candidates.
                 break
+
+        for event in self.events[len(self.events)-count:]:
+            slots[event["name"]] = event["value"]
 
         return slots
 
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -1534,10 +1534,6 @@ class TestFormValidationAction(FormValidationAction):
             return {
                 "slot2": "validated_value",
             }
-        if tracker.get_slot("slot1") == "validated_value":
-            return {
-                "slot2": "slot1 == validated_value",
-            }
         return {
             "slot2": None,
         }
@@ -1585,7 +1581,7 @@ async def test_form_validation_action():
     assert not warnings
     assert events == [
         SlotSet("slot1", "validated_value"),
-        SlotSet("slot2", "slot1 == validated_value"),
+        SlotSet("slot2", None),
     ]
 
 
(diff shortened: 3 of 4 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @TyDunn will take a look at it as soon as possible ✨

TyDunn:
Hi @cmanovit. Can you please add a description to this PR? Why do you want to preserve the order of slots to be validated (rather than reversed)?

rasa_sdk/forms.py
```diff
@@ -864,6 +864,9 @@ async def validate(
 
             if isinstance(validation_output, dict):
                 slots.update(validation_output)
+                # for sequential consistency, also update tracker
+                # to make changes visible to subsequent validate_{slot_name}
+                tracker.slots.update(validation_output)
```
wochinge, line 869: I think this should work on a copy of the tracker to avoid changing the original tracker for other functions

rasa_sdk/interfaces.py
```diff
@@ -247,17 +247,21 @@ def slots_to_validate(self) -> Dict[Text, Any]:
...
             # at the end of the tracker events.
             if event["event"] == "slot":
-                slots[event["name"]] = event["value"]
+                count += 1
             else:
                 # Stop as soon as there is another event type as this means that we
                 # checked all potential slot candidates.
                 break
 
+        for event in self.events[len(self.events)-count:]:
```
wochinge, line 262: the change makes sense to me - just wondering if we can find a more elegant way to do this (e.g. using `enumerate`). Especially the name `count` in my opinion hides the fact that we are dealing with indexes here.

wochinge (commented):
I think the change makes sense 💯 Could you please add tests as well?

rasa_sdk/forms.py
wochinge: mhm, actually we're changing the tracker directly [here](https://github.com/RasaHQ/rasa-sdk/blob/86f41e094df9565d915d5a597dce6878ba304c1f/rasa_sdk/forms.py#L701) as well - so I think it's fine to keep it like it is

wochinge: we need to get rid of this line then though: https://github.com/RasaHQ/rasa-sdk/blob/86f41e094df9565d915d5a597dce6878ba304c1f/rasa_sdk/forms.py#L704

rasa_sdk/interfaces.py
cmanovit: "count" is indeed the count of slot events but I can understand your concern given it is later used to compute index (but same would go for len() which can often be used in index calculation also). I agreed enumerate would be more elegant but I was not confident enough to make such change.

What if we simply iterate through self.events entirely in its original order at the first place? I am not familiar with the internals enough to gauge the impact of the loss performance but the code should be easier to read and free from the assumption of slot events being last.

rasa_sdk/forms.py
cmanovit: Could you please suggest how to best add a test for this? I could only think of making a full Rasa example which I am not sure if it is reasonable.

Did you actually mean to remove the line "tracker.add_slots(validation_events)" as it should be redundant now? - https://github.com/RasaHQ/rasa-sdk/blob/86f41e094df9565d915d5a597dce6878ba304c1f/rasa_sdk/forms.py#L699

ancalita: Hi @cmanovit I would encourage you to add tests in `[tests.test_forms`](https://github.com/RasaHQ/rasa-sdk/blob/86f41e094df9565d915d5a597dce6878ba304c1f/tests/test_forms.py) module. There are plenty of examples that you can follow to some extent, all seem to have these steps:
- create a `TestForm` class inheriting from `FormValidationAction`and initialize it as `form`
- create an instance of Tracker and an instance of `CollectingDispatcher` which can then be used as arguments to `form.run()`
- then I guess it would make sense to assert that `tracker.slots` match the expected updated slots after validation?

Also I'm not sure L699 should be removed, the `tracker.add_slots` method updates not only `tracker.slots` but also `tracker.events`. If you do remove that line, you'd need to update tracker.events in `FormValidationAction.validate` too.

ancalita (commented):
Could you also please sign the [Contributor License Agreement](https://cla-assistant.io/RasaHQ/rasa-sdk?pullRequest=441) as prompted by the CLA assistant?

rasa_sdk/forms.py
cmanovit: Hi @ancalita Thank you for the tip! It seems there were already tests which check the order of the returned slots and they had failed due to the change of behavior. So, I added another commit to update those tests accordingly and they are passing again.

ancalita:
Hi @cmanovit could you please reformat your files using `make formatter` and then push the changes again?

rasa_sdk/forms.py
cmanovit: If this change makes sense, then I suppose it should be similarly applied to extract_custom_slots. So, I went ahead and made the change and also added a new test for it.

tests/test_forms.py
```diff
@@ -1534,6 +1534,10 @@ def validate_slot2(
             return {
                 "slot2": "validated_value",
             }
+        if tracker.get_slot("slot1") == "validated_value":
```
ancalita, line 1537: Why are these new lines necessary?

tests/test_forms.py
```diff
@@ -1580,8 +1584,8 @@ async def test_form_validation_action():
 
     assert not warnings
     assert events == [
-        SlotSet("slot2", None),
         SlotSet("slot1", "validated_value"),
+        SlotSet("slot2", "slot1 == validated_value"),
```
ancalita, line 1588: I agree that the order of slots needs to be changed in the list after the change you proposed - however the test should work with `SlotSet("slot2", None)`

tests/test_forms.py
```diff
@@ -1805,6 +1809,101 @@ async def validate_my_slot(
...
+@pytest.mark.parametrize(
+    "my_required_slots, extracted_values, asserted_events",
+    [
+        (
+            ["state", "city"],
+            {"state": "california", "city": "san francisco"},
+            [["state", "CA"], ["city", "San Francisco"], [REQUESTED_SLOT, None]],
+        ),
+        (
+            ["city", "state"],
```
ancalita, line 1821: I see that in the first test the order of `my_required_slots` is first "state", then "city". In the 2nd test, the order is reversed, was this intentional? if so, why?

tests/test_forms.py
```diff
@@ -1805,6 +1809,101 @@ async def validate_my_slot(
...
+            return {"city": slot_value}
+
+    form = TestFormValidationWithCustomSlots()
+
+    # tracker with active form
+    tracker = Tracker(
+        "default",
+        {},
+        {},
+        [],
```
ancalita, line 1894: I'm also curious why events is empty here?

tests/test_forms.py
```diff
@@ -1805,6 +1809,101 @@ async def validate_my_slot(
...
+        False,
+        None,
+        {"name": "some_form", "is_interrupted": False, "rejected": False},
+        "action_listen",
+    )
+
+    dispatcher = CollectingDispatcher()
+    events = await form.run(dispatcher=dispatcher, tracker=tracker, domain={})
```
ancalita, line 1902: Shouldn't the `domain` dict contain "forms" dict with the custom slots?

rasa_sdk/interfaces.py
wochinge: > What if we simply iterate through self.events entirely in its original order at the first place?

Mhm, there could be a lot of events from the conversation (I guess something in the range from 0 to 100) while we can be sure that the information we need (the `SlotSet` events) are at the end of this list.

> free from the assumption of slot events being last.

That assumption is actually crucial. Otherwise there is no way to distinguish whether the slots are set by the `FormAction` within Rasa Open Source or whether they were set earlier in the conversation.

tests/test_forms.py
cmanovit: Yes, this is intentional. This is to test that, with commit eaf8d86, the extracted value of "city" can subsequently influence the extraction of "state" - that is, knowing the city is san francisco should allow us to infer that the state is california without requiring to see the term in user's message.

tests/test_forms.py
cmanovit: I copied this test from the one right before it and modified it for my purpose and might have overlooked many details. That test also has empty events list. I suppose it is because we are testing extraction of custom slots here and previous events may not be too relevant.

tests/test_forms.py
cmanovit: I think the document indicates that custom slots which do not use predefined mappings should not be listed in the domain. (https://rasa.com/docs/rasa/forms/#custom-slot-mappings) I am not entirely sure but it seems the existing tests on custom slots do not set this neither.

rasa_sdk/interfaces.py
cmanovit: Thanks @wochinge for the explanation.

tests/test_forms.py
ancalita: You should create a separate test for this use-case because as far as I can see the scenario when `slot2` would be None is never tested.

cmanovit: Okay, I have reverted it back as what I wanted to check there had already been included in the new test that I added.

ancalita:
Hi @cmanovit apologies for the delay on my end - could you also please add a changelog entry [here](https://github.com/RasaHQ/rasa-sdk/tree/main/changelog), this would classify as `improvement`. Thank you 💯

ancalita (approved):

PR #502, opened 2021-07-19
501 drop py36 support

### LINKED ISSUE #500: Update Websockets to 9.1 to fix CVE-2021-33880 when we drop support for Python 3.6
**Rasa version**: NA

**Rasa SDK version**: rasa/rasa-sdk:2.8.0 

**Python version**: NA

**Operating system** (windows, osx, ...): NA

**Issue**:

We need to update websockets to version 9.1 to fix [CVE-2021-33880](https://nvd.nist.gov/vuln/detail/CVE-2021-33880) when we[ drop support for Python 3.6](https://github.com/RasaHQ/rasa/issues/8569).

We can't do this right now because websockets 9.1 requires sanic >= 21 which requires dropping support for python 3.6.

```

```

```

```

**Content of configuration file (config.yml)** (if relevant):
```yml

```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

**Contents of actions.py** (if  relevant):
```python

```

### LINKED ISSUE #501: Drop support for Python 3.6
**Description of Problem**:
Python 3.6 end of life is [Dec 23rd 2021](https://endoflife.date/python). Our next major release could be an opportunity to drop support

**Definition of Done**:
- [ ] CI workflows are updated
- [ ] Documentation is updated
- [ ] Internal documentation is updated
- [ ] sanic is updated to v21.6.0
- [ ] branch / tag is created to be used for sanic update in rasa oss

### LINKED ISSUE #503: Upgrade Sanic to v21.6.0
**Description of Problem**:
* We are dropping support for python 3.6 #502  and can thus upgrade sanic
* that new version fixes #500 
* the new version also comes with more type annotations

**Overview of the Solution**:
There are two potential ways forward:
1. sticking with the `pytest-sanic` and waiting for them to upgrade their sanic and websocket dependencies, and then upgrading ours.
2. dropping `pytest-sanic` dependency as the new version of sanic introduces `sanic-testing` which seems to be similar to what `pytest-sanic` does, e.g. providing test clients. I have tried that path briefly, but all the test requests would return 500 (internal server error) and I didn't find the problem quickly so I am creating this issue here.

**Definition of Done**:
- [x] sanic is upgraded to ^21.6.0
- [x] websockets is upgraded to ^9.1
- [x] sanic-testing is added as a dependency
- [ ] maybe: pytest-sanic is removed as a dependency -> nope still needed, otherwise only ~100 out of ~150 tests run
- [x] tests are working
- [x] create issues more specific steps and learnings from this issue for the sanic upgrades in rasa-x and rasa Oss

twerkmeister:
**Proposed changes**:
- fixes #501 
- fixes #500 by updating sanic to > 21.0.0
- fixes #503 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.black]
 line-length = 88
-target-version = [ "py36", "py37", "py38",]
+target-version = [ "py37", "py38",]
 exclude = "((.eggs | .git | .mypy_cache | .pytest_cache | build | dist))"
 
 [tool.poetry]
@@ -16,7 +16,7 @@ maintainers = [ "Tom Bocklisch <tom@rasa.com>",]
 homepage = "https://rasa.com"
 repository = "https://github.com/rasahq/rasa-sdk"
 documentation = "https://rasa.com/docs"
-classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Programming Language :: Python", "Programming Language :: Python :: 3.6", "Programming Language :: Python :: 3.7", "Topic :: Software Development :: Libraries",]
+classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Programming Language :: Python", "Programming Language :: Python :: 3.7", "Programming Language :: Python :: 3.8", "Topic :: Software Development :: Libraries",]
 keywords = [ "nlp", "machine-learning", "machine-learning-library", "bot", "bots", "botkit", "rasa conversational-agents", "conversational-ai", "chatbot", "chatbot-framework", "bot-framework",]
 include = [ "LICENSE.txt", "README.md",]
 readme = "README.md"
@@ -63,9 +63,9 @@ name = "Miscellaneous internal changes"
 showcontent = false
 
 [tool.poetry.dependencies]
-python = ">=3.6,<3.9"
+python = ">=3.7,<3.9"
 coloredlogs = ">=10,<16"
-sanic = ">=19.12.2,<21.0.0"
+sanic = "^21.6.0"
 sanic-cors = "^0.10.0"
 requests = ">=2.23.0,<2.26.0"
 typing-extensions = "^3.7.4"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,9 +66,9 @@ showcontent = false
 python = ">=3.7,<3.9"
 coloredlogs = ">=10,<16"
 sanic = "^21.6.0"
-sanic-cors = "^0.10.0"
 requests = ">=2.23.0,<2.26.0"
 typing-extensions = "^3.7.4"
+Sanic-Cors = "^1.0.0"
 
 [tool.poetry.dev-dependencies]
 pytest-cov = "^2.12.1"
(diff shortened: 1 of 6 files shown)


twerkmeister:
currently blocked by https://github.com/yunstanford/pytest-sanic/issues/55

twerkmeister:
why sanic cors needed upgrading too: https://community.sanicframework.org/t/handle-request-exception-after-upgrading-from-20-9-1-to-21-3/835

rasa_sdk/endpoint.py
```diff
@@ -127,13 +127,14 @@ async def actions(_) -> HTTPResponse:
 
 def run(
     action_package_name: Union[Text, types.ModuleType],
-    port: Union[Text, int] = DEFAULT_SERVER_PORT,
+    port: int = DEFAULT_SERVER_PORT,
```
m-vdb, line 130: if we were allowing `Text` here (likely from an env var?) should we cast the value below with `int(port)` instead of only changing the type?

twerkmeister: I had walked back the code inside rasa sdk and saw that it is only called with the args.port which is forced to be int. Also took another look rasa oss and rasa x also don't call this. Using `int(port)` might be safer for user code though

m-vdb: ok sounds good then, if the user does not call this directly, I don't think it's necessary to typecast it

m-vdb (approved):
💯

pyproject.toml
```diff
@@ -63,12 +63,12 @@ name = "Miscellaneous internal changes"
...
 [tool.poetry.dependencies]
-python = ">=3.6,<3.9"
+python = ">=3.7,<3.9"
 coloredlogs = ">=10,<16"
-sanic = ">=19.12.2,<21.0.0"
-sanic-cors = "^0.10.0"
+sanic = "^21.6.0"
 requests = ">=2.23.0,<2.26.0"
 typing-extensions = "^3.7.4"
+Sanic-Cors = "^1.0.0"
```
m-vdb, line 71: could you keep this lowercase though? it seems that both would work

pyproject.toml
```diff
@@ -63,12 +63,12 @@ name = "Miscellaneous internal changes"
 showcontent = false
 
 [tool.poetry.dependencies]
-python = ">=3.6,<3.9"
+python = ">=3.7,<3.9"
 coloredlogs = ">=10,<16"
-sanic = ">=19.12.2,<21.0.0"
-sanic-cors = "^0.10.0"
+sanic = "^21.6.0"
```
m-vdb, line 68: oh actually this conflicts with what is currently defined for Rasa Open Source, do we also need to bump the priority here?

And should we allow for ">=21.6.0,<22.0.0" to simplify upgrades?

m-vdb (changes requested):
Approved too fast, see latest comments :)

pyproject.toml
twerkmeister: That's what I got runnning `poetry add sanic-cors=^1.0.0` I also find it very odd. Yet, pypi also lists the package with those capatalized letters. https://pypi.org/project/Sanic-Cors/

pyproject.toml
twerkmeister: ^21.6.0 actually means exactly that if I am not mistaken. Minor + micro updates are allowed, but not major ones

twerkmeister: yes we should also upgrade it in rasa oss, I have created an issue for that: https://github.com/RasaHQ/rasa/issues/9315

rasa_sdk/endpoint.py
twerkmeister: At least we don't seem to call it directly in rasa oss and x. But users could be doing that, that's what I meant to say. Usually we would probably want a deprecation warning here - but given that it's such a small change we might just do it and get rid of a bit of clutter?

m-vdb: no I think it's fine. If it's not in our guidelines to have user call this method directly, I think it's fine

pyproject.toml
m-vdb: 🤷🏻 magic

m-vdb (approved):
@twerkmeister sounds good! I let you merge the PR when you want

PR #506, opened 2021-07-22
Do not fail if a Python package distribution exists in the target package index

tczekajlo:
**Proposed changes**:
- Do not fail if a Python package distribution exists in the target package index

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -300,10 +300,11 @@ jobs:
       run: poetry build
 
     - name: Publish to PyPI 📦
-      uses: pypa/gh-action-pypi-publish@37e305e7413032d8422456179fee28fac7d25187
+      uses: pypa/gh-action-pypi-publish@bea5cda687c2b79989126d589ef4411bedce0195
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -300,10 +300,11 @@ jobs:
       run: poetry build
 
     - name: Publish to PyPI 📦
-      uses: pypa/gh-action-pypi-publish@37e305e7413032d8422456179fee28fac7d25187
+      uses: pypa/gh-action-pypi-publish@bea5cda687c2b79989126d589ef4411bedce0195
```
joejuzl, line 303: Is there a reason we use the commit hashes rather than releases?

joejuzl (changes requested):

.github/workflows/continuous-integration.yml
tczekajlo: Yes, we use hashes for GH action for security reasons, we wanna be sure that someone didn't re-tag, or did something odd.

joejuzl (approved):
🚀

PR #549, opened 2021-09-29
Slot mappings

alwx:
**Proposed changes**:
- ...

### DIFF
--- a/CHANGELOG.mdx
+++ b/CHANGELOG.mdx
@@ -117,14 +117,14 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
   want to extract an additional slot, add the slot's name to the list of `required_slots`
   and add a method `extract_<slot name>` to your action:
 
-  ```python
+```python
   from typing import Text, Dict, Any, List, Optional
 
   from rasa_sdk.forms import (
-      FormValidationAction,
+      ValidationAction,
   )
 
-  class FormWithSlotExtractions(FormValidationAction):
+  class FormWithSlotExtractions(ValidationAction):
       def name(self) -> Text:
           return "some_form"
 
@@ -231,13 +231,13 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
 
   Example:
 
-  ```python
+```python
   from typing import Text, Any, Dict
-  from rasa_sdk import FormValidationAction, Tracker
+  from rasa_sdk import ValidationAction, Tracker
   from rasa_sdk.types import DomainDict
   from rasa_sdk.executor import CollectingDispatcher
 
-  class MyFormValidationAction(FormValidationAction):
+  class MyFormValidationAction(ValidationAction):
       def name(self) -> Text:
           return "some_form"
 
--- a/rasa_sdk/forms.py
+++ b/rasa_sdk/forms.py
@@ -44,7 +44,9 @@ class SlotMapping(Enum):
 
     @staticmethod
     def intent_is_desired(
-        mapping: Dict[Text, Any], tracker: "Tracker", domain: "DomainDict",
+        mapping: Dict[Text, Any],
+        tracker: "Tracker",
+        domain: "DomainDict",
     ) -> bool:
         """Check whether user intent matches intent conditions"""
         mapping_intents = SlotMapping.to_list(mapping.get("intent", []))
@@ -65,7 +67,8 @@ class SlotMapping(Enum):
 
     @staticmethod
     def entity_is_desired(
-        mapping: Dict[Text, Any], tracker: "Tracker",
+        mapping: Dict[Text, Any],
+        tracker: "Tracker",
     ) -> bool:
         """Checks whether slot should be filled by an entity in the input or not.
 
@@ -77,16 +80,18 @@ class SlotMapping(Enum):
             True, if slot should be filled, false otherwise.
         """
         slot_fulfils_entity_mapping = False
-        extracted_entities = tracker.latest_message.entities
+        extracted_entities = tracker.latest_message.get("entities", [])
 
         for entity in extracted_entities:
             if (
-                    mapping.get("entity") == entity["entity"]
-                    and mapping.get("role") == entity.get("role")
-                    and mapping.get("group") == entity.get("group")
+                mapping.get("entity") == entity["entity"]
+                and mapping.get("role") == entity.get("role")
+                and mapping.get("group") == entity.get("group")
             ):
                 matching_values = tracker.get_latest_entity_values(
-                    mapping.get("entity"), mapping.get("role"), mapping.get("group"),
+                    mapping.get("entity"),
+                    mapping.get("role"),
+                    mapping.get("group"),
                 )
                 slot_fulfils_entity_mapping = matching_values is not None
                 break
@@ -95,7 +100,9 @@ class SlotMapping(Enum):
 
     @staticmethod
     def _get_ignored_intents(
-        mapping: Dict[Text, Any], domain: "DomainDict", active_loop_name: Text,
+        mapping: Dict[Text, Any],
+        domain: "DomainDict",
+        active_loop_name: Text,
     ) -> List[Text]:
         mapping_conditions = mapping.get("conditions")
         active_loop_match = False
@@ -262,7 +269,11 @@ class FormAction(Action):
             SlotMapping.to_list(not_intent),
         )
 
-        return {"type": str(SlotMapping.FROM_TEXT), "intent": intent, "not_intent": not_intent}
+        return {
+            "type": str(SlotMapping.FROM_TEXT),
+            "intent": intent,
+            "not_intent": not_intent,
+        }
 
     # noinspection PyMethodMayBeStatic
     def slot_mappings(self) -> Dict[Text, Union[Dict, List[Dict[Text, Any]]]]:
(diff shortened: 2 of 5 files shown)


rasa_sdk/forms.py
```diff
@@ -656,29 +559,8 @@ async def run(
...
-
-        for i in range(1, len(mappings)):
-            if entity_type != mappings[i].get("entity"):
-                return None
-
-        return entity_type
-
 
-class FormValidationAction(Action, ABC):
+class ValidationAction(Action, ABC):
```
ancalita, line 563: I haven't yet looked in detail in all the PR diff, however thought of flagging that this part of the issue is incomplete. This class needs to be also redesigned to work for slots that can be extracted in a custom way outside forms, at the moment it only serves forms. This is what I was referring when asking about the naming convention when the action should set slots outside forms (e.g., the `validate_slots` prefix I used in the rasa default action). Let me know if you prefer to jump on a quick call today/tomorrow to discuss this?

alwx: Let's jump on a call early tomorrow.

rasa_sdk/forms.py
```diff
@@ -49,134 +52,17 @@ def required_slots(tracker: "Tracker") -> List[Text]:
             "A form must implement required slots that it has to fill"
```
wochinge, line 52: Can't we deprecate the entire `FormAction` as it was only used with the old `FormPolicy` which was dropped with Rasa Open Source 3.0?

rasa_sdk/forms.py
```diff
@@ -656,29 +555,259 @@ async def run(
     def __str__(self) -> Text:
         return f"FormAction('{self.name()}')"
 
-    def _get_entity_type_of_slot_to_fill(
+
+class SeparateValidationAction(Action, ABC):
```
wochinge, line 559: can we keep the old name / create an alias with a deprecation so it doesn't break for users when updating?

CHANGELOG.mdx
```diff
@@ -117,14 +117,14 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
...
+```python
   from typing import Text, Dict, Any, List, Optional
 
   from rasa_sdk.forms import (
-      FormValidationAction,
+      ValidationAction,
   )
 
-  class FormWithSlotExtractions(FormValidationAction):
+  class FormWithSlotExtractions(ValidationAction):
```
ancalita, line 127: Is this update to superclass necessary since we're preserving `FormValidationAction`?

rasa_sdk/forms.py
```diff
@@ -21,6 +23,8 @@
 
 LOOP_INTERRUPTED_KEY = "is_interrupted"
 
+VALIDATE_GLOBAL_SLOT_MAPPINGS_NAME = "validate_global_slot_mappings"
```
ancalita, line 26: In this case, should I update the action name searched by default action in rasa-oss to this one?

rasa_sdk/forms.py
ancalita: +1, considering the warning message in this `FormAction` constructor, this was deprecated in 2.0 so it can be safely removed now, alongside tests? Also, then the `SlotMapping` code can also be removed.

rasa_sdk/forms.py
```diff
@@ -656,46 +557,41 @@ async def run(
...
+    def name(self) -> Text:
+        """Unique identifier of this simple action."""
+        if self.global_slot_mappings:
+            return VALIDATE_GLOBAL_SLOT_MAPPINGS_NAME
 
-class FormValidationAction(Action, ABC):
-    """A helper class for slot validations and extractions of custom slots."""
+        raise NotImplementedError("An action must implement a name")
 
     def form_name(self) -> Text:
```
ancalita, line 571: I think this method makes sense only for `FormValidationAction` but not in the parent class, should this be moved there instead?

rasa_sdk/forms.py
```diff
@@ -656,46 +557,41 @@ async def run(
...
         """Returns the form's name."""
         return self.name().replace("validate_", "", 1)
 
+    @property
+    def global_slot_mappings(self) -> bool:
```
ancalita, line 576: Would this method have to be overriden by the user in case it can be used outside forms - just to make sure I account for this in the docs?

rasa_sdk/forms.py
```diff
@@ -705,7 +601,31 @@ async def run(
         # Validation events include events for extracted slots
         return validation_events
 
-    async def extract_custom_slots(
+    async def required_slots(
```
ancalita, line 604: Why does this method have to be `async`? Also can we not remove it  and use `domain_slots` directly? Or is it preserved to continue for the user to override it as in 2.0?

tests/test_forms.py
```diff
@@ -1552,6 +1653,206 @@ async def validate_slot3(
...
+        {},
+        [
+            UserUttered(
+                "My name is Emily.",
+                parse_data={
+                    "intent": {"name": "inform", "confidence": 1.0},
+                    "entities": [{"entity": "name", "value": "Emily"}],
+                },
+            ),
+            SlotSet("slot1", "Emily"),
```
ancalita, line 1759: Why is this event required?

tests/test_forms.py
```diff
@@ -1552,6 +1653,206 @@ async def validate_slot3(
...
+    }
+
+    dispatcher = CollectingDispatcher()
+    with pytest.warns(None) as warnings:
+        events = await validation_action.run(
+            dispatcher=dispatcher, tracker=tracker, domain=domain,
+        )
+
+    assert not warnings
+    assert events[0] == SlotSet("slot1", "Emily")
```
ancalita, line 1790: If `slot1` can only be set in an active loop and the tracker doesn't have an active loop running, shouldn't actually the output be an empty list?

ancalita (commented):
Thank you for the quick submission 💯 
Left a few questions and suggestions.

rasa_sdk/forms.py
alwx: Yep the idea was to not change the signature of this function to make the migration easier. If we change it from `async` to `non-async` then the users would need to introduce loops in their custom implementations of `required_slots` when migrating from 2.0 to 3.0

CHANGELOG.mdx
alwx: Just updated it.
`FormValidationAction` needs to be used for forms; `ValidationAction` for slots outside forms.

tests/test_forms.py
alwx: To indicate the the slot was set. This test doesn't run anything to do the slot extraction (like any other `rasa-sdk` test), it just uses the events from the current tracker and the domain to check if the extracted slot needs to be validated.

rasa_sdk/forms.py
```diff
@@ -916,3 +749,99 @@ async def next_requested_slot(
...
+        return self.name().replace("validate_", "", 1)
+
+    def domain_slots(self, domain: "DomainDict") -> List[Text]:
+        """Returns slots which were mapped in the domain.
+
+        Args:
+            domain: The current domain.
+
+        Returns:
+            Slot names mapped in the domain.
```
ancalita, line 842:
```suggestion
            Slot names which should be filled by the form. By default it
            returns the slot names which are listed for this form in the domain
            and use predefined mappings.
```

rasa_sdk/forms.py
```diff
@@ -705,7 +589,29 @@ async def run(
...
+        Args:
+            domain_slots: Names of slots of this form which were mapped in
+                the domain.
+            dispatcher: the dispatcher which is used to
+                send messages back to the user.
+            tracker: the conversation tracker for the current user.
+            domain: the bot's domain.
+
+        Returns:
+            A dictionary of slot names.
```
ancalita, line 610:
```suggestion
            A list of slot names.
```

rasa_sdk/forms.py
```diff
@@ -848,9 +668,15 @@ async def validate(
...
             `SlotSet` events for every validated slot.
         """
+        slots_to_validate = await self.required_slots(
+            self.domain_slots(domain), dispatcher, tracker, domain
+        )
         slots: Dict[Text, Any] = tracker.slots_to_validate()
 
         for slot_name, slot_value in list(slots.items()):
+            if slot_name not in slots_to_validate:
+                continue
```
ancalita, line 678: It took me a while, but I think I understand why in `test_validation_action_outside_forms_with_form_active_loop`, the output events is not an empty list, because the slot persists in `slots` un-updated. Shouldn't the functionality be not to add an identical SlotSet to the tracker, but to return an empty list? So in this line, probably the slot needs to be removed from `slots` 🤔

CHANGELOG.mdx
alwx: Oh got it, you're right

rasa_sdk/forms.py
```diff
@@ -692,10 +572,14 @@ async def run(
...
+        )
         tracker.add_slots(extraction_events)
 
-        validation_events = await self.validate(dispatcher, tracker, domain)
+        validation_events = await self.get_validation_events(
+            dispatcher, tracker, domain
+        )
         tracker.add_slots(validation_events)
 
         next_slot = await self.next_requested_slot(dispatcher, tracker, domain)
```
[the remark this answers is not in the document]
alwx: It doesn't seem like it will always be `None` because `required_slots` can still be overridden in a specific implementation of `ValidationAction`.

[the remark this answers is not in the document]
alwx: Done

PR #591, opened 2021-11-11
Add custom docs theme with og:image meta tag

maggo:
Port of https://github.com/RasaHQ/rasa/pull/10090

**Proposed changes**:
- Adds `og:image` file and property using a custom docs theme


melindaloubser1:
@maggo Is this PR still relevant?

PR #836, opened 2022-08-25
2.8.x allow external logging formatters

raoulvm:
**Proposed changes**:
To be able to use structured logging, there has to be a dedicated method to define the logger for the Rasa SDK server:
- introduce an environment variable `RASA_LOGGER_PYMODULE` to store the python package name of the custom logger
- catch the previously uncaught exceptions from route handlers, so the exception output is also logged and not written to stderr in plain text.

Background: we use GrayLog with JSON extractor for all our servers, and 

### DIFF
--- a/rasa_sdk/__main__.py
+++ b/rasa_sdk/__main__.py
@@ -26,14 +26,11 @@ def main_from_args(args):
             invalidate_caches()
             import_module(name=logger_path)
             # update root level
-            logging.getLogger(__name__).warning(f"set root level to {args.loglevel}")
+            logging.getLogger(__name__).debug(f"set root level to {args.loglevel}")
             logging.basicConfig(level=args.loglevel, force=True)
-
-            print("===CATCHME======")
         except Exception as e:
             default_logging(args)
             logging.getLogger(__name__).exception("custom logger import failed", exc_info=e)
-            print("===CAUGHT++++++++")
     else:
         default_logging(args)
 
--- a/rasa_sdk/__main__.py
+++ b/rasa_sdk/__main__.py
@@ -1,18 +1,37 @@
 import logging
+from importlib import import_module, invalidate_caches
+from os import environ
 
 from rasa_sdk import utils
-from rasa_sdk.endpoint import create_argument_parser, run
 from rasa_sdk.constants import APPLICATION_ROOT_LOGGER_NAME
+from rasa_sdk.endpoint import create_argument_parser, run
+
+RASA_LOGGER_PYMODULE = "RASA_LOGGER_PYMODULE"
+
+
+def default_logging(args):
+    utils.configure_colored_logging(args.loglevel)
+    utils.configure_file_logging(
+        logging.getLogger(APPLICATION_ROOT_LOGGER_NAME), args.log_file, args.loglevel
+    )
 
 
 def main_from_args(args):
     """Run with arguments."""
     logging.getLogger("matplotlib").setLevel(logging.WARN)
 
-    utils.configure_colored_logging(args.loglevel)
-    utils.configure_file_logging(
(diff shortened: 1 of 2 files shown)


raoulvm:
@tmbo Here the PR I mentioned during the call on Tuesday.

tmbo:
@raoulvm I've taken a closer look and ideally I'd like to aim at a unified solution across our different product parts (SDK, Rasa Open Source, Rasa Enterprise).

A standard method for the configuration of the logging system is to use a file based configuration rather than relying on code / a custom environment variable. 

I'd propose the following:
* add an additional command line argument / startup parameter / environment variable to the different products that allows you to specify a logging configuration file, e.g. you can specify the configuration in the following ways
  * `rasa run --logging_configuration=my_logging_configration.yml` or alternatively
  *  `RASA_LOGGING_CONFIGURATION=my_logging_configuration.yml rasa run`
* define the logging of the different products using https://docs.python.org/3/library/logging.config.html#dictionary-schema-details, e.g. a configuration would look like this:
  ```yml
  formatters:
    brief:
      format: '%(levelname)-8s: %(name)-15s: %(message)s'
    precise:
      format: '%(asctime)s %(name)-15s %(levelname)-8s %(message)s'
  filters:
    allow_foo:
      name: foo
  handlers:
    console:
      class : logging.StreamHandler
      formatter: brief
      level   : INFO
      stream  : ext://sys.stdout
      filters: [allow_foo]
    file:
      class : logging.handlers.RotatingFileHandler
      formatter: precise
      filename: logconfig.log
      maxBytes: 1024
      backupCount: 3
    debugfile:
      class : logging.FileHandler
      formatter: precise
      filename: logconfig-detail.log
      mode: a
  loggers:
    foo:
      level : ERROR
      handlers: [debugfile]
    spam:
      level : CRITICAL
      handlers: [debugfile]
      propagate: no
    bar.baz:
      level: WARNING
  root:
    level     : DEBUG
    handlers  : [console, file]
  ```

Does this solve your configuration needs for logging?

tmbo:
@raoulvm I've moved your changes regarding the exception handling to a separate PR https://github.com/RasaHQ/rasa-sdk/pull/844

raoulvm:
> @raoulvm I've taken a closer look and ideally I'd like to aim at a unified solution across our different product parts (SDK, Rasa Open Source, Rasa Enterprise).
> 
> A standard method for the configuration of the logging system is to use a file based configuration rather than relying on code / a custom environment variable.
> 
> I'd propose the following:
> 
> * add an additional command line argument / startup parameter / environment variable to the different products that allows you to specify a logging configuration file, e.g. you can specify the configuration in the following ways
>   
>   * `rasa run --logging_configuration=my_logging_configration.yml` or alternatively
>   * `RASA_LOGGING_CONFIGURATION=my_logging_configuration.yml rasa run`
> * define the logging of the different products using https://docs.python.org/3/library/logging.config.html#dictionary-schema-details, e.g. a configuration would look like this:
>   ```yaml
>   formatters:
>     brief:
>     ....
>   ```
> 
> Does this solve your configuration needs for logging?

Hi @tmbo this looks great, different though, so I have to look deeper into that. Our current implementation uses a modified logger class, and that is not configurable using the standard `dictConfig()` so I have to look into re-writing that to move functionality from there to the handler or filter classes. Should be possible though.

ancalita (commented):
Hi @raoulvm Checking in if the solution @tmbo released solved your needs and if we can close this PR?

tmbo:
well, we still need to implement the logging part :D (we only shipped the exception handler part of the PR)

ancalita:
@raoulvm I've implemented the changes (that @tmbo mentioned above) in [this PR](https://github.com/RasaHQ/rasa-sdk/pull/877). Could you please take a look and validate if this implementation satisfies your requirements before I merge it imminently? 🙏🏻

PR #844, opened 2022-08-30
created catch all for exceptions to handle them properly

tmbo:
**Proposed changes**:
- handle exceptions thrown in custom actions
- return a 500 response code as well as information about the error 
- compatible change since we already by default returned a 500 (sanic default). but exception was not properly logged to logger

### DIFF
--- a/changelog/836.improvement.md
+++ b/changelog/836.improvement.md
@@ -1,0 +1,1 @@
+Return a json response when encountering action execution exceptions. Log exceptions using logger rather than sanic default print.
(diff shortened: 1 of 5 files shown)


ancalita (approved):
LGTM 💯

raoulvm:
Great, thank you @tmbo

raoulvm:
Oh, I just noticed 2.8.9 was already released. Will this be 2.8.10, and what is the release schedule?

tmbo:
@raoulvm we cut that slightly to early but we already prepared a follow up release: https://github.com/RasaHQ/rasa-sdk/pull/846. the change is part of 2.8.10

We always release patches when we need them / on demand (only minors/majors are scheduled releases for us).

tmbo:
2.8.10 is available on pypi https://pypi.org/project/rasa-sdk/2.8.10/

PR #900, opened 2022-11-24
Return default value as github release body

rasa-aadlv:
- Fix publishing empty release notes


scripts/publish_gh_release_notes.py
```diff
@@ -57,6 +57,9 @@ def parse_changelog(tag_name: Text) -> Text:
         if consuming_version:
             version_lines.append(line)
 
+    # return default value as github release body
+    if not version_lines:
+        return "No significant changes."
```
m-vdb, line 62: so actually, I looked at the changelog file, and it seems that [it's broken](https://github.com/RasaHQ/rasa-sdk/blob/main/CHANGELOG.mdx#321---2022-08-30-no-significant-changes). The "No significant changes" seems to be added to the heading. So it seems that something is wrong with the changelog generation using towncrier. It looks like we addressed the issue on Rasa Open Source with [this PR](https://github.com/RasaHQ/rasa/pull/11529/files)

m-vdb (changes requested):

scripts/publish_gh_release_notes.py
rasa-aadlv: That's all correct. However we should keep this as a fallback to ensure that our GitHub release doesn't break only because of missing notes in the body.

rasa-jmac:
@m-vdb the fix for this is now as it was fixed in the OSS repo

m-vdb (approved):
thanks @rasa-aadlv @rasa-jmac 🙏🏻

PR #926, opened 2023-01-25
Domain caching

raoulvm:
**Proposed changes**:
- Add an environment variable to point to a domain definition endpoint (e.g. rasa.server:5005/domain) to _actively_ pull domain data on first action execution request (e.g. after server (re)start).  
  If domain is sent from rasa OSS server with each request (default setting), no endpoint is required.
- See [PR11966](https://github.com/RasaHQ/rasa/pull/11966) of the Rasa Server to disable sending the domain with each request to the action server.
- format is
`DOMAIN_ENDPOINT=http[s]://server.tld:port/path[?param=value[&...]]` e.g.  
`DOMAIN_ENDPOINT=http://rasa.local:5005/token=thisismysecret`


rasa_sdk/endpoint.py
```diff
@@ -4,23 +4,280 @@
...
+        subpath = subpath[1:]
+    return url + subpath
+
+
+local_domain = dict()
+domain_endpoint = None
+
+logger.debug(f"Domain Endpoint is {DOMAIN_ENDPOINT}")
+
+if DOMAIN_ENDPOINT:
```
tmbo, line 223: i am a little worried about the service dependency this introduces. with the change added, a rasa SDK deployment now depends on rasa open source. previously, only rasa open source dependent on the SDK. with this change the dependency becomes bidirectional where there was previously only a unidirectional dependency.

raoulvm: But only if you enable this feature. If you don't - or you have another compatible domain endpoint somewhere else, you don't have the dependency.

raoulvm: Also, it never tries to contact the domain endpoint if the Rasa server sends a domain with the request (default behavior), so you need to actively switch that off in Rasa server with the env var introduced in [Rasa PR11966](https://github.com/RasaHQ/rasa/pull/11966)

raoulvm:
Has been solved differently by Rasa with https://github.com/RasaHQ/rasa/pull/12013 (planned for 3.4.3)

Customers are still free to implement this part on their own and switch off all domain data in the requests if they want to.

PR #930, opened 2023-02-01
[ATO-712] Port deflate custom action request to 3.3.x

radovanZRasa:
**Proposed changes**:
Add capability to handle deflated body of a HTTP request.

### DIFF
--- a/changelog/930.improvement.md
+++ b/changelog/930.improvement.md
@@ -1,0 +1,3 @@
+Added the ability to decompress deflate encoded
+json payloads for the webhook endpoint of the 
+action server.
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -2,6 +2,8 @@ import argparse
 import logging
 import os
 import types
+import zlib
+import json
 from typing import List, Text, Union, Optional, Any
 from ssl import SSLContext
 
@@ -90,7 +92,13 @@ def create_app(
     @app.post("/webhook")
     async def webhook(request: Request) -> HTTPResponse:
         """Webhook to retrieve action calls."""
-        action_call = request.json
+        if request.headers.get("Content-Encoding") == "deflate":
+            # Decompress the request data using zlib
+            decompressed_data = zlib.decompress(request.body)
+            # Load the JSON data from the decompressed request data
+            action_call = json.loads(decompressed_data)
+        else:
+            action_call = request.json
         if action_call is None:
             body = {"error": "Invalid body request"}
             return response.json(body, status=400)
--- a/rasa_sdk/endpoint.py
+++ b/rasa_sdk/endpoint.py
@@ -14,7 +14,7 @@ from sanic_cors import CORS
 
 from rasa_sdk import utils
 from rasa_sdk.cli.arguments import add_endpoint_arguments
-from rasa_sdk.constants import DEFAULT_SERVER_PORT
+from rasa_sdk.constants import DEFAULT_SERVER_PORT, DEFLATED_REQUEST_BODY_HEADER
 from rasa_sdk.executor import ActionExecutor
 from rasa_sdk.interfaces import ActionExecutionRejection, ActionNotFoundException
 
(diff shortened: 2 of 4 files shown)


rasa_sdk/endpoint.py
```diff
@@ -90,7 +92,13 @@ async def health(_) -> HTTPResponse:
     @app.post("/webhook")
     async def webhook(request: Request) -> HTTPResponse:
         """Webhook to retrieve action calls."""
-        action_call = request.json
+        if request.headers.get("Content-Encoding") == "deflate":
```
ancalita, line 95: Are there other types of content-encoding that we're sending, or this is the only one?
I'd also extract `deflate` to a constant.

rasa_sdk/endpoint.py
```diff
@@ -90,7 +92,13 @@ async def health(_) -> HTTPResponse:
...
+            # Decompress the request data using zlib
+            decompressed_data = zlib.decompress(request.body)
+            # Load the JSON data from the decompressed request data
+            action_call = json.loads(decompressed_data)
```
ancalita, line 99: Are there any exceptions that could be thrown during decompression that we need to handle?

rasa_sdk/endpoint.py
```diff
@@ -90,7 +92,13 @@ async def health(_) -> HTTPResponse:
     @app.post("/webhook")
     async def webhook(request: Request) -> HTTPResponse:
         """Webhook to retrieve action calls."""
-        action_call = request.json
+        if request.headers.get("Content-Encoding") == "deflate":
+            # Decompress the request data using zlib
+            decompressed_data = zlib.decompress(request.body)
```
ancalita, line 97: Shouldn't this be `request.json` instead of `request.body`?

changelog/930.improvement.md
```diff
@@ -0,0 +1,3 @@
+Added the ability to decompress deflate encoded
```
ancalita, line 1: Could we also add a unit test for this change?

rasa_sdk/endpoint.py
radovanZRasa: This is the only one.

rasa_sdk/endpoint.py
radovanZRasa: Not really. This is non-recoverable error per-request.

rasa_sdk/endpoint.py
radovanZRasa: Compressed body will not be stored in json property.

rasa_sdk/endpoint.py
m-vdb: but that's a good point. I think both decompression errors + json decoding error be caught, logged and a 400 response should be returned (similar to line 104 below)

m-vdb (changes requested):

rasa_sdk/endpoint.py
```diff
@@ -90,7 +93,26 @@ async def health(_) -> HTTPResponse:
...
+                logger.debug(e)
+                body = {"error": "Error while decompressing body of HTTP request"}
+                return response.json(body, status=400)
+
+            try:
+                action_call = json.loads(decompressed_data)
+            except JSONDecodeError as e:
+                logger.debug(e)
+                body = {"error": e.msg}
+                return response.json(body, status=400)
```
m-vdb, line 112:
```suggestion
            try:
                # Decompress the request data using zlib
                decompressed_data = zlib.decompress(request.body)
                # Load the JSON data from the decompressed request data
                action_call = json.loads(decompressed_data)
             except zlib.error as e:
                logger.debug(e)
                body = {"error": "Error while decompressing body of HTTP request"}
                return response.json(body, status=400)
            except JSONDecodeError as e:
                logger.debug(e)
                body = {"error": e.msg}
                return response.json(body, status=400)
```
(indentation seems messed up, sorry)


The last block in your code duplicates some of the work I think. Additionally, you'll need to define `action_call = None` before the `if` at line 96, otherwise there's at least a branch of code where the `action_call` variable is not defined

radovanZRasa: Oh, I missed that try-catch duplication. Thx for spotting that!

I believe we do not need to set action_call = None, cause there is an if-else branch which will cover initialization of action_call.

m-vdb: you're right -- must have been confused

m-vdb (approved):

raoulvm:
How will this be configured? Because our production environment **does not support compressed request bodies**.
So this needs to be configurable in `endpoints` or somewhere else.

radovanZRasa:
> How will this be configured? Because our production environment **does not support compressed request bodies**. So this needs to be configurable in `endpoints` or somewhere else.

This would be configured on Rasa OSS side. There you will be able to toggle if it should send compressed body of an HTTP request for custom actions.
Rasa SDK will automatically detect if body of an HTTP request for custom actions is compressed or not and act accordingly.

PR #933, opened 2023-02-09
Replace `set-output` command because its deprecated

miraai:
**Proposed changes**:
- Replace set-output command because its deprecated


rasa-jmac (commented):
I think we should actually push this onto `3.4.x` so that we don't end up with release builds failing when this is deprecated, what do you think?

PR #949, opened 2023-03-15
Setup ruff

m-vdb:
**Proposed changes**:
- Relates to https://github.com/RasaHQ/rasa/pull/12135
- Replace flake8 with `ruff`: https://beta.ruff.rs/docs/


scripts/lint_python_docstrings.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Lint docstrings only against the the diff to avoid too many errors.
+# Check only production code. Ignore other other errors which are captured by `lint`
```
radovanZRasa, line 4: > Check only production code.
What does the production code mean?

rasa_sdk/__init__.py
```diff
@@ -1,15 +1,14 @@
 import logging
 
+import rasa_sdk.cli
 import rasa_sdk.version
+from rasa_sdk.interfaces import Tracker, Action, ActionExecutionRejection  # noqa: F401
```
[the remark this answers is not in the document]
m-vdb: I am not sure what you mean here. Can you explain?

setup.cfg
```diff
@@ -1,38 +0,0 @@
-# pytest pycodestyle configuration
-[tool:pytest]
-# Function starting with the following pattern are considered for test cases.
-python_functions=test_
-
-# pytest pycodestyle configuration
-codestyle_max_line_length = 88
-codestyle_ignore = E302 W503 E203 E501 E265 E402 E251 E211
```
ancalita, line 8: I think these configurations for `pytest` are missing from `pyproject`

ancalita (approved):

setup.cfg
m-vdb: but the thing is that these were invalid. They are pycodestyle configuration that should be for flake8, not pytest 🤔 I think these were kind of legacy config that we kept along and never updated.

ancalita: Gotcha 👍🏻

scripts/lint_python_docstrings.sh
radovanZRasa: > Check only runtime code 

sounds more clear.

rasa_sdk/__init__.py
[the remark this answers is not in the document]
radovanZRasa: I do not see that these imports are used by the code below. That is why I think that some other parts of the imported code require these imports in `__init__.py`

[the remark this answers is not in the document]
m-vdb: it's because they're part of the public "API" of the SDK. We want people to be able to do `from rasa_sdk import ...`, for ease of use. I wouldn't change it

radovanZRasa (approved):

PR #1012, opened 2023-07-25

Urkem:
**Proposed changes**:
- ...

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -627,7 +627,7 @@ multidict = ">=5.0,<6.0"
 sanic-routing = ">=0.7,<1.0"
 ujson = {version = ">=1.35", markers = "sys_platform != \"win32\" and implementation_name == \"cpython\""}
 uvloop = {version = ">=0.5.3", markers = "sys_platform != \"win32\" and implementation_name == \"cpython\""}
-websockets = ">=10.0,<11.0"
+websockets = ">=10.0"
 
 [package.extras]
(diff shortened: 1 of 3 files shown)


poetry.lock
```diff
@@ -627,7 +627,7 @@ multidict = ">=5.0,<6.0"
 sanic-routing = ">=0.7,<1.0"
 ujson = {version = ">=1.35", markers = "sys_platform != \"win32\" and implementation_name == \"cpython\""}
 uvloop = {version = ">=0.5.3", markers = "sys_platform != \"win32\" and implementation_name == \"cpython\""}
-websockets = ">=10.0,<11.0"
```
ancalita, line 630: I think we need to keep this constraint on.

poetry.lock
```diff
@@ -682,7 +682,7 @@ doc = ["Sphinx", "sphinx-rtd-theme"]
 name = "setuptools"
 version = "65.6.3"
```
ancalita, line 683: Considering poetry lock already has an entry for a version that matches the constraint in pyproject, do we absolutely need to explicitly add setuptools to pyproject?

ancalita: Nm, I saw the change in category from dev to main too late

rasa-jmac (approved):

poetry.lock
Urkem: Seems like this was a constraint never ported to 3.4.x 😞 it's present on 3.5.x and 3.6.x tho

ancalita: Yeah I vaguely remembered the fix on this branch was to directly amend `poetry.lock` although it's a brittle solution, since the lock file gets updated whenever we make changes to pyproject.

Urkem: I've added it manually and to pyproject.toml

ancalita (approved):

PR #1013, opened 2023-07-25

Urkem:
**Proposed changes**:
- ...


poetry.lock
```diff
@@ -682,9 +682,9 @@ doc = ["Sphinx", "sphinx-rtd-theme"]
 
 [[package]]
 name = "setuptools"
-version = "67.0.0"
```
ancalita, line 685: Similar here, why do we need to downgrade to this version in particular if setuptools is already at a higher version?

Urkem: Well the security report attached to the ticked specified we are using an lower version 🤔 , probably because we don't specify `setuptools` in `pyproject.toml` and the scan does not check the `lock` file...., Anyway I've removed the `^` from the version

ancalita (approved):

PR #1014, opened 2023-07-25

Urkem:
**Proposed changes**:
- ...


Tawakalt (approved):

poetry.lock
```diff
@@ -1151,14 +1152,14 @@ doc = ["Sphinx", "sphinx-rtd-theme"]
 
 [[package]]
 name = "setuptools"
-version = "67.0.0"
```
ancalita, line 1154: Similar question here, why don't we stay upgraded to the current version?
I also see the category change from dev to main - that probably means setuptools is promoted from a dev dependency to a main dependency, I'd prefer to update the constraint in pyproject to have it at the same version as currently in the lock file.

Urkem: reverted the change to min instead of `^`

PR #1051, opened 2023-12-01
Update pep440-version-utils

tabergma:
**Proposed changes**:
Update pep440-version-utils


tabergma:
Should this PR target `main` or `3.7.x`?

pyproject.toml
```diff
@@ -79,13 +79,13 @@ coloredlogs = ">=10,<16"
 sanic = "^21.12.0"
 typing-extensions = ">=4.1.1,<5.0.0"
 Sanic-Cors = "^2.0.0"
-prompt-toolkit = "^3.0,<3.0.29"
```
ancalita, line 82: This update has to be validated with rasa-private CLI commands (like `rasa interactive` or `rasa shell`), because they had introduced a breaking change in `3.0.30` which I'm not sure they addressed.

pyproject.toml
```diff
@@ -79,13 +79,13 @@ coloredlogs = ">=10,<16"
...
+prompt-toolkit = ">=3,<4"
 "ruamel.yaml" = ">=0.16.5,<0.18.0"
 websockets = ">=10.0,<12.0"
 pluggy = "^1.0.0"
 opentelemetry-api = "~1.15.0"
 opentelemetry-sdk = "~1.15.0"
-opentelemetry-exporter-jaeger = "~1.15.0"
```
ancalita, line 88: Two points here, if we're updating this dependency, all the other open telemetry dependencies have to be on the same minor. In addition, we have to align this dependency version with the identical dependencies in rasa-plus.

ancalita:
> Should this PR target `main` or `3.7.x`?

Housekeeping PRs should always target `main`, only dependency vulnerability fixes should target release branches.

pyproject.toml
tabergma: Thanks @ancalita for pointing it out. As it seems to be much more work than expected, I would close this PR and create an issue for it. We are currently focusing on a project that needs to be done by end of the year.

PR #1057, opened 2024-01-03
[ATO-887] Upgrade Sanic and Sanic-testing

Tawakalt:
**Proposed changes**:
- Fixes [ATO-887](https://rasahq.atlassian.net/browse/ATO-887)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,7 +99,7 @@ toml = "^0.10.0"
 pep440-version-utils = "^0.3.0"
 semantic_version = "^2.8.5"
 mypy = "^1.5"
-sanic-testing = "^22.3.0, <22.9.0"
+sanic-testing = "^22.12"
 
 [tool.ruff.pydocstyle]
 convention = "google"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ select = [ "D", "E", "F", "W", "RUF",]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 coloredlogs = ">=10,<16"
-sanic = "^21.12.0"
+sanic = ">=22.9.1,<22.12"
 typing-extensions = ">=4.1.1,<5.0.0"
 Sanic-Cors = "^2.0.0"
 prompt-toolkit = "^3.0,<3.0.29"
--- a/poetry.lock
+++ b/poetry.lock
@@ -1397,31 +1397,32 @@ files = [
 
 [[package]]
 name = "sanic"
-version = "21.12.2"
+version = "22.9.1"
 description = "A web server and web framework that's written to go fast. Build fast. Run fast."
 category = "main"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "sanic-21.12.2-py3-none-any.whl", hash = "sha256:ef4edddfba46f2f8728400470f84deb91d9e5fc21cf2531512acf70638699620"},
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -76,7 +76,7 @@ select = [ "D", "E", "F", "W", "RUF",]
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
 coloredlogs = ">=10,<16"
-sanic = "^21.12.0"
+sanic = ">=22.9.1,<22.12"
```
ancalita, line 79: I would double-check what's the latest LTS of sanic we could upgrade to, I can't remember if `22.0.0` is one of them.

tests/test_endpoint.py
```diff
@@ -110,3 +105,10 @@ def test_server_webhook_custom_action_encoded_data_returns_200():
 
     assert events == [SlotSet("test", "bar")]
     assert response.status == 200
+
+
+# ENSURE THIS IS ALWAYS THE LAST TEST FOR OTHER TESTS TO RUN
+# for some reason, sys.exit is actually called here
```
ancalita, line 111: This is actually expected given this line that gets called as a result of this test: https://github.com/RasaHQ/rasa-sdk/blob/c1fec2a333dd01145e363ab4ff739eed88448f3a/rasa_sdk/executor.py#L266 
I would investigate more on why this breaks the pytest command if placed anywhere else in the test module, maybe other users of Sanic have encountered this as well?

pyproject.toml
Tawakalt: You're right.
Our LTS [options](https://sanic.dev/en/organization/policies.html#long-term-support-v-interim-releases) are 23.12 or 22.12

tests/test_endpoint.py
Tawakalt: Yes, I did some investigation and couldn't find anything helpful.
It appeared the line was actually being executed by pytest, so any other thing afterwards isn't being run.

Not sure why it appeared with the updated Sanic version though and I couldn't find anything linking it to the issue.

pyproject.toml
ancalita: 23.12 seems released on 31-Dec, could be coming with undetected bugs? I'll let you decide which version is best.

Tawakalt: Yeah.
We have till Dec to use 22.12, so I'll use that instead.

tests/test_endpoint.py
ancalita: I'm wondering if it's not because of the incongruent version of sanic-testing with sanic? I'd test this again once you upgrade to sanic 22.12, that's the version you used for sanic-testing.

Tawakalt: I was hoping that was the case too and tried it after the upgrade but it still persists.

Tawakalt: There should be a package that manages the ordering, but I'm not sure the effort is worth it for now.

ancalita: Yeah definitely not worth the effort.

ancalita: I'd just replace the last line comment to say `call to sys.exit() terminates pytest process` or smth along those lines.

ancalita: Do we know if all the tests in the suite run actually, maybe we actually need to disable this and investigate later?

ancalita: Ah just checked the CI, all looks good.

Tawakalt: Comment updated.

ancalita (approved):

PR #1075, opened 2024-02-13
[ATO-2105] Implement functionality to create custom spans

Tawakalt:
**Proposed changes**:
- Fixes [ATO-2105](https://rasahq.atlassian.net/browse/ATO-2105)

- [X] added some tests for the functionality
- [X] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa-sdk/tree/main/changelog) for instructions)


tests/tracing/test_trace_provider.py
```diff
@@ -0,0 +1,19 @@
...
+def test_anonymization_pipeline_provider_is_singleton() -> None:
+    trace_provider_1 = TraceProvider()
+    trace_provider_2 = TraceProvider()
+
+    assert trace_provider_1 is trace_provider_2
+    assert trace_provider_1.tracer is trace_provider_2.tracer
+
+
+def test_trace_provider() -> None:
+    trace_provider = TraceProvider()
```
ancalita, line 14: Let's add an assertion here that `trace_provider.get_tracer` is None.

rasa_sdk/tracing/trace_provider.py
```diff
@@ -0,0 +1,23 @@
+from typing import Optional
+from rasa_sdk.utils import Singleton
+from opentelemetry.trace import Tracer
+
+
+class TraceProvider(metaclass=Singleton):
```
ancalita, line 6: I would rename this component not to be confused with OpenTelemetry's TracerProvider.
Let's rename it to `TracerRegister` or something like that.

tests/tracing/test_trace_provider.py
```diff
@@ -0,0 +1,19 @@
+from rasa_sdk.tracing.trace_provider import TraceProvider
+from opentelemetry import trace
+
+
+def test_anonymization_pipeline_provider_is_singleton() -> None:
```
ancalita, line 5: Test name should be updated.

tests/tracing/instrumentation/test_action_executor.py
```diff
@@ -56,3 +60,30 @@ async def test_tracing_action_executor_run(
     assert captured_span.name == "MockActionExecutor.run"
 
     assert captured_span.attributes == expected
+
+
+@pytest.mark.asyncio
+async def test_instrument_action_executor_run_registers_tracer(
```
ancalita, line 66: Why is this marked as async? We're not using any async methods or functions in this test?

Tawakalt: I removed the async call and didn't update.

ancalita: Oh the action executor's `run` method is async, that's why this is required, i'd add a docstring or comment to describe this.

rasa_sdk/tracing/trace_provider.py
```diff
@@ -0,0 +1,23 @@
...
+    """Represents a provider for tracer."""
+
+    tracer: Optional[Tracer] = None
```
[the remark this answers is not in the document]
Tawakalt: Hmmm...since this will always refer to the same data, what's the point of the dictionary? 🤔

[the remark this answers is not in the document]
Tawakalt: Or do we wanna have multiple tracers based on the instrumented class module?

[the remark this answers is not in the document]
ancalita: > have multiple tracers based on the instrumented class module?

Yup in case of that. We should add docstrings to document this too.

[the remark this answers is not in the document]
Tawakalt: This will make things a bit complicated for the users having to know which tracer to use.

Having multiple tracer options based on the instrumented class module won't really change much. Only these tags are dependent on the tracer used. The positioning of the span doesn't change.
Screenshot 2024-02-13 at 17 09 11

Also instrumented methods in the code may or may not be called by the custom action. That's why I chose the ActionExecutor since it's a common factor.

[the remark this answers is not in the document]
ancalita: Ok, in that case we need to make it explicit in the name and docstring that this trace register is only for the ActionExecutor's tracer e.g. `ActionExecutorTracerRegister`.

Tawakalt:
> Great work 💯 In addition to the below comments, how did you obtain that view in Jaeger? I'd suggest to add a new test custom action that does the same steps you took to get that custom span in a different test case.

It will require an integration test to achieve this.

I wrote a custom action like this and triggered it
```
class ValidatePredefinedSlots(ValidationAction):
    def validate_location(
        self,
        slot_value: Any,
        dispatcher: CollectingDispatcher,
        tracker: Tracker,
        domain: DomainDict,
    ) -> Dict[Text, Any]:
        """Validate location value."""
        tracer = TracerRegister().get_tracer()
        with tracer.start_as_current_span("my_very_own_custom_span"):
            if isinstance(slot_value, str):
                # validation succeeded, capitalize the value of the "location" slot
                dispatcher.utter_message("Nice! I've always wanted to visit ", slot_value.capitalize())
                return {"location": slot_value.capitalize()}
            else:
                # validation failed, set this slot to None
                return {"location": None}

tests/tracing/instrumentation/test_action_executor.py
Tawakalt: yes, but i didn't call the method in the test.
I've updated the method to not be async.

ancalita (approved):
🎉 
I'd also make Sanchari aware of this integration test case need to be added to the list for the action server testing.

PR #1104, opened 2024-05-15
Return 449 status code if the Domain is not stored or missing in the …

aleksandarmijat:
**Proposed changes**:
- Checks the domain in the payload and return 449 status code if the domain is missing (both stored and in the payload) or stale (digest mismatch).

### DIFF
--- a/rasa_sdk/executor.py
+++ b/rasa_sdk/executor.py
@@ -166,8 +166,8 @@ class ActionExecutor:
         self.actions: Dict[Text, Callable] = {}
         self._modules: Dict[Text, TimestampModule] = {}
         self._loaded: Set[Type[Action]] = set()
-        self.domain: Dict[Text, Any] = None
-        self.domain_digest: Text = None
+        self.domain: Optional[Dict[Text, Any]] = None
+        self.domain_digest: Optional[Text] = None
 
     def register_action(self, action: Union[Type[Action], Action]) -> None:
         if inspect.isclass(action):
@@ -386,7 +386,7 @@ class ActionExecutor:
                 # we won't append this to validated events -> will be ignored
         return validated
 
-    def is_domain_digest_valid(self, domain_digest: Text) -> bool:
+    def is_domain_digest_valid(self, domain_digest: Optional[Text]) -> bool:
         """Check if the domain_digest is valid
 
         If the domain_digest is empty or different from the one provided, it is invalid.
@@ -401,7 +401,7 @@ class ActionExecutor:
 
     def update_and_return_domain(
         self, payload: Dict[Text, Any], action_name: Text
-    ) -> Dict[Text, Any]:
+    ) -> Optional[Dict[Text, Any]]:
         """Validate the digest, store the domain if available, and return the domain.
 
         This method validates the domain digest from the payload.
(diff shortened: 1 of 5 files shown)


rasa_sdk/executor.py
```diff
@@ -162,6 +166,8 @@ def __init__(self) -> None:
         self.actions: Dict[Text, Callable] = {}
         self._modules: Dict[Text, TimestampModule] = {}
         self._loaded: Set[Type[Action]] = set()
+        self.domain: Dict[Text, Any] = {}
```
radovanZRasa, line 169: Have you considered to use `None` instead of empty dict and empty string below?

rasa_sdk/interfaces.py
```diff
@@ -384,3 +384,14 @@ def __init__(self, action_name: Text, message: Optional[Text] = None) -> None:
...
     def __str__(self) -> Text:
         return self.message
+
+
+class ActionMissingDomainException(Exception):
+    """Raising this exception when the domain is missing."""
+
+    def __init__(self, action_name: Text, message: Optional[Text] = None) -> None:
+        self.action_name = action_name
+        self.message = message or f"Domain context is missing."
```
radovanZRasa, line 394: No need for formatted string. You can ditch the `f` at the start of string.

rasa_sdk/executor.py
```diff
@@ -380,6 +386,49 @@ def validate_events(events: List[Dict[Text, Any]], action_name: Text):
...
+
+        Args:
+            domain_digest: latest value provided to compare the current value with.
+
+        Returns:
+            True if the domain_digest is valid, False otherwise.
+        """
+        return bool(self.domain_digest) and self.domain_digest == domain_digest
+
+    def get_domain(
```
radovanZRasa, line 402: Word `get` suggests that the method is not doing any mutation of the data while in fact it is. It is mutating the state of the parent object.
I would suggest to explore different name like: `upsert_domain` or `update_domain` etc.

rasa_sdk/executor.py
aleksandarmijat: I thought it was more safe to have these variables be of the type it is expected. It's also how variables above are defined. You want to differentiate a case where if it's `None` it means that there is no Domain information, and if it is `{}` and `""` the Domain information is there, it's just empty Domain, right?

rasa_sdk/interfaces.py
aleksandarmijat: Definitely. CI caught it as well.

rasa_sdk/executor.py
aleksandarmijat: You are right. Though this method is not just "upserting" the Domain, it's also returning the result. Should the naming match that behavior as well? What would you name it in that case?

rasa_sdk/executor.py
radovanZRasa: Yea, I was thinking about clearly differentiating case when there is a domain (whether it may be empty or not) and when there is no domain info at all.

aleksandarmijat: Noted, makes sense to me, I'll update and test it.

aleksandarmijat: I had to update the typing to use `Optional[...]` so that `None` value can be accepted. This resulted in additional changes so that the code can pass the Code Quality check, like removing `ActionCall` as typing and using `Dict[Text, Any]` (which is the correct type for `action_call`.

rasa_sdk/executor.py
aleksandarmijat: I updated the method name to `update_and_return_domain` to clearly state that we are making changes to the domain fields of the parent object AND also returning the domain.

Also updated the docstring to describe this in a more precise manner.

aleksandarmijat:
[Spike](https://rasahq.atlassian.net/browse/ATO-2354) got approved, closing this PR, will be used as the inspiration for the implementation.

PR #1124, opened 2024-07-09
Build multi-platform docker images on release

radovanZRasa:
Rearrange build of Docker images in GitHub workflows for easier management:
* Add new workflow to build image when code is merged to `main`
* Build Docker dev image with pr### (ex. pr1234) tag on pull requests, devs should use this image if they need it in `rasa` 
* Move build of Docker images and PyPi package when tag is pushed to a new workflow
* Add separate workflow to cleanup dev docker image created with pr### tag when pull request is merged. Devs who are working on `rasa` should switch to the `latest` image once PR on `rasa-sdk` is merged.


radovanZRasa:
The check which is hanging should be replaced in the repo's settings but I would prefer we first do a code review before that is done.

Dockerfile
```diff
@@ -26,10 +26,10 @@ RUN apt-get update -qq \
 
 # install poetry
 # keep this in sync with the version in pyproject.toml and Dockerfile
-ENV POETRY_VERSION 1.8.2
+ENV POETRY_VERSION=1.8.2
```
sanchariGr, line 29: Is it possible to use poetry version as an env var from [here](https://github.com/RasaHQ/rasa-sdk/blob/main/scripts/poetry-version.sh) instead of hardcoding it? Makes it easier when upgrading poetry.

.github/workflows/pr-cleanup.yml
```diff
@@ -0,0 +1,51 @@
+name: CI On Pull Request Closure
```
sanchariGr, line 1: This workflow was purposely not added. This would mess up the workflow in rasa-private which always fetches the latest image from aws sdk repo [here](https://github.com/RasaHQ/rasa-private/blob/main/.github/workflows/continous-integration.yml#L253). 
Also I have configured the AWS repo to clean up images every 2 days so there wont be older images left behind in the aws repo.

.github/workflows/pr-merged.yml
```diff
@@ -1,11 +1,9 @@
...
 
 on:
-  schedule:
-    # Run cron job at 8AM Monday to Sunday.
-    - cron: '0 8 * * *'
-  workflow_dispatch:
-  pull_request:
+  push:
+    branches:
+      - main
```
sanchariGr, line 6: Also means there wont be latest docker images available in AWS for the rasa-private workflow [here](https://github.com/RasaHQ/rasa-private/blob/main/.github/workflows/continous-integration.yml#L253) . We don't work on sdk code too much so this push to main will rarely happen.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,161 @@
+name: Release artifacts
+
+on:
+  push:
+    tags:
+      - "*"
```
sanchariGr, line 6: Can we add a `workflow_dispatch` here, if this workflow fails then we would be able to trigger that manually with a workflow dispatch.

sanchariGr (changes requested):

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,161 @@
+name: Release artifacts
```
sanchariGr, line 1: Love this splitting up of the release specific workflow but in this repo we still manually have to push tags. Would you suggest adding the auto tag feature in the release script and the tag workflow like [this](https://github.com/RasaHQ/rasa-private/blob/main/.github/workflows/tag-release.yml) to this repo as a follow up after this is merged?

Dockerfile
radovanZRasa: Done.

.github/workflows/pr-cleanup.yml
radovanZRasa: Oh, yes, I forgot about the cleanup rule we have in AWS ECR.

.github/workflows/pr-merged.yml
radovanZRasa: What is on `main` is what should be in the `latest` docker image built. 
I added a job which will build Docker image and tag it with pr number.
The AWS rule will not delete this image because it is tagged.
Devs which are working on `rasa-private` repo can use PR tagged Docker image (which will not be deleted) unless PR is closed. Once a PR on Rasa SDK is merged dev can switch to `latest` Docker image

Previous workflow would have race-condition which would overwrite `latest` image in case there were two or more PRs opened on the SDK.

.github/workflows/release-artifacts.yml
radovanZRasa: Added.

.github/workflows/release-artifacts.yml
radovanZRasa: I was going for inlining what we have in other repositories.

sanchariGr (approved):
🔥

.github/workflows/release-artifacts.yml
radovanZRasa: So far we have manually pushed tags so I did not want to make any changes there.
We could add similar script here as well. I would do that in separate PR.

PR #1209, opened 2025-02-10
Update grpcio to resolve dependendency issues in rasa-private

sanchariGr:
**Proposed changes**:
- Fixes the following issue
`Because rasa-sdk (3.12.0.dev1) depends on grpcio (>=1.67.1,<1.68.0)
   and pymilvus (2.4.1) depends on grpcio (>=1.49.1,<=1.60.0), rasa-sdk (3.12.0.dev1) is incompatible with pymilvus (2.4.1).
  And because no versions of pymilvus match >2.4.1,<2.4.2, rasa-sdk (3.12.0.dev1) is incompatible with pymilvus (>=2.4.1,<2.4.2).
  So, because rasa-pro depends on both pymilvus (>=2.4.1,<2.4.2) and rasa-sdk (3.12.0.dev1), version solving failed.`
- Fixes also fixes issue with `setup-qemu-action` discussed [here](https://rasa-hq.slack.com/archives/C061J0LGHU0/p1739369378917929)


sanchariGr:
> Oh seems like we need to regenerate the file [here](https://github.com/RasaHQ/rasa-sdk/tree/main/proto) - I'm not sure if we have this documented somewhere in the README, if not, @radovanZRasa could you please share the steps to do so?

@ancalita couldnt find a README so until @radovanZRasa replies, I'll create a dev release for sdk from this branch and use that in rasa-private.

PR #1220, opened 2025-04-02
update poetry to 2.1.2 on 3.9.x

sanchariGr:
**Proposed changes**:
- update poetry to 2.1.2 on 3.9.x


.github/poetry_version.txt
```diff
@@ -1,2 +1,2 @@
 # The poetry version is stored in a separate file due to the https://github.com/python-poetry/poetry/issues/3316
-poetry-version=1.8.2
+poetry-version=2.0.0
```
ancalita, line 2: I think poetry already released a new minor - should we upgrade to `2.1` or whatever latest micro there is?

PR #1222, opened 2025-04-11
Change AWS region for DEV docker images to eu

rasa-aadlv:
- Rasa Pro CI cluster will be in eu-west-1, so wee need to keep the images in the same zone to avoid additional fees
- Change AWS region for DEV docker images to eu-west-1


sanchariGr:
https://github.com/RasaHQ/rasa-sdk/blob/main/.github/workflows/continuous-integration.yml#L14 this needs to be updated.

sanchariGr (approved):

PR #1314, opened 2026-04-09
[PF-1131] Enable response streaming from custom actions

ancalita:
**Proposed changes**:
- Support added for response streaming from custom actions via direct custom action execution and gRPC only.
- https://rasahq.atlassian.net/browse/PF-1131


rasa_sdk/grpc_server.py
```diff
@@ -179,6 +182,181 @@ def convert_metadata_to_multidict(
...
+        with tracer.start_as_current_span(span_name, context=tracing_context) as span:
+            check_version_compatibility(request.version)
+            if self.auto_reload:
+                self.executor.reload()
+
+            action_call = MessageToDict(request, preserving_proto_field_name=True)
+            action_name = action_call.get("next_action", "")
+            response_id = uuid.uuid4().hex
+            sink: asyncio.Queue = asyncio.Queue()
+
```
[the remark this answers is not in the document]
ancalita: The queue is intentionally unbounded. Each custom action invocation is finite; the total number of enqueued items is bounded by the chunks the action emits. Adding maxsize would risk deadlocking if the consumer is paused while the action awaits put() inside a finally block or the auto-close guard.
In addition, gRPC applies backpressure at the generator level. grpc.aio iterates the async generator by calling __anext__() and internally awaits the write to the HTTP/2 send buffer before requesting the next message. When the client's receive window is exhausted, grpc.aio delays the next __anext__() call. The generator is therefore idle during that period. However — and this is the crux — the generator being idle does not suspend _run(), because _run() is a separate asyncio.Task on the same event loop. So _run() can advance during that window.

rasa_sdk/grpc_server.py
```diff
@@ -179,6 +182,184 @@ def convert_metadata_to_multidict(
...
+                    if result is None:
+                        # next_action was absent; executor.run() returned without
+                        # placing stream_done, so we must do it here.
+                        await sink.put({"event": "stream_done", "result": None})
+                except (
+                    ActionExecutionRejection,
+                    ActionNotFoundException,
+                    ActionMissingDomainException,
+                ) as exc:
+                    await sink.put({"event": "stream_error", "exception": exc})
```
Tawakalt: Isn't this and line 261 a duplicate? `ActionExecutor.run` already does this in line 746.

rasa_sdk/grpc_server.py
```diff
@@ -179,6 +182,184 @@ def convert_metadata_to_multidict(
...
+                            chunk=ParseDict(
+                                {
+                                    k: chunk[k]
+                                    for k in (
+                                        "text",
+                                        "image",
+                                        "custom",
+                                        "attachment",
+                                        "buttons",
+                                        "elements",
```
Tawakalt, line 297: `CollectingDispatcher.stream_chunk` accepts `**kwargs`, which is not included in this chunk. Anything else in `**kwargs` is omitted from the stream. Is that intentional?

rasa_sdk/executor.py
```diff
@@ -65,6 +129,88 @@ def utter_message(
...
+            payload["image"] = image
+        if json_message:
+            payload["custom"] = json_message
+        if attachment is not None:
+            payload["attachment"] = attachment
+        if buttons:
+            payload["buttons"] = buttons
+        if elements:
+            payload["elements"] = elements
+        payload.update(kwargs)
```
Tawakalt, line 213: The docstring says template/response aren’t supported, but `kwargs` is merged directly into the payload, so template/response, can still be passed indirectly.
Do we need to do something about that.

rasa_sdk/executor.py
```diff
@@ -178,7 +324,26 @@ class ActionExecutorRunResult(BaseModel):
...
+    execution path:
+
+    * :meth:`run` — unary path.  No sink is attached to the dispatcher, so
+      streaming methods buffer their chunks and replay them as one or more
+      :meth:`~CollectingDispatcher.utter_message` calls when streaming ends.
+      All responses are returned together in the result payload.
+    * :meth:`run_streaming` — streaming path.  The executor injects an
+          :class:`asyncio.Queue` as the dispatcher's sink so that chunks
+          produced by :meth:`~CollectingDispatcher.stream_chunk` are forwarded
+          immediately, before the action finishes.
```
Tawakalt:
```suggestion
      immediately, before the action finishes.
```

rasa_sdk/executor.py
```diff
@@ -178,7 +324,26 @@ class ActionExecutorRunResult(BaseModel):
...
+    :class:`CollectingDispatcher`; the executor selects the appropriate
+    execution path:
+
+    * :meth:`run` — unary path.  No sink is attached to the dispatcher, so
+      streaming methods buffer their chunks and replay them as one or more
+      :meth:`~CollectingDispatcher.utter_message` calls when streaming ends.
+      All responses are returned together in the result payload.
+    * :meth:`run_streaming` — streaming path.  The executor injects an
+          :class:`asyncio.Queue` as the dispatcher's sink so that chunks
+          produced by :meth:`~CollectingDispatcher.stream_chunk` are forwarded
```
Tawakalt:
```suggestion
      produced by :meth:`~CollectingDispatcher.stream_chunk` are forwarded
```

rasa_sdk/executor.py
```diff
@@ -178,7 +324,26 @@ class ActionExecutorRunResult(BaseModel):
...
+    Action authors always write the same code using
+    :class:`CollectingDispatcher`; the executor selects the appropriate
+    execution path:
+
+    * :meth:`run` — unary path.  No sink is attached to the dispatcher, so
+      streaming methods buffer their chunks and replay them as one or more
+      :meth:`~CollectingDispatcher.utter_message` calls when streaming ends.
+      All responses are returned together in the result payload.
+    * :meth:`run_streaming` — streaming path.  The executor injects an
+          :class:`asyncio.Queue` as the dispatcher's sink so that chunks
```
Tawakalt:
```suggestion
      :class:`asyncio.Queue` as the dispatcher's sink so that chunks
```

tests/test_endpoint.py
```diff
@@ -41,7 +41,7 @@ def test_server_list_actions_returns_200(
 
     # Then the server should return a list of actions
     assert response.status == 200
-    assert len(response.json) == 9
+    assert len(response.json) == 11
     print(response.json)
```
Tawakalt:
```suggestion
```

rasa_sdk/executor.py
ancalita: Yes, you're right - I'll update to raise ValueError if response keys are provided to `CollectingDispatcher.stream_chunk()`.

rasa_sdk/grpc_server.py
ancalita: Yes, this is intentional, the Chunk protobuf type needs to indicate fields precisely. Without a particular use-case for arbitrary field extras, I thought it's best to stick to known parameters in rasa-sdk CollectingDispatcher API. I will make a note to document this.

rasa_sdk/grpc_server.py
ancalita: Good catch, we shouldn't add duplicate `stream_error` events, so I'll update to make sure this doesn't happen.

rasa_sdk/executor.py
ancalita: Addressed: [26d696f](https://github.com/RasaHQ/rasa-sdk/pull/1314/commits/26d696f4d72addbd8f7087c68499512ded5c4651)

rasa_sdk/grpc_server.py
ancalita: Addressed: [26d696f](https://github.com/RasaHQ/rasa-sdk/pull/1314/commits/26d696f4d72addbd8f7087c68499512ded5c4651)

Tawakalt (approved):
🚀

PR #1353, opened 2026-06-23
[INFRA-1967] Workflow security hardening

rasa-aadlv:
[JIRA](https://rasahq.atlassian.net/browse/INFRA-1967)

- Pin github3.py to 4.0.1 and pep440-version-utils to 1.2.0
in the release-atifacts-publish-release job, which runs with
contents: write and access to RASASDK_GITHUB_TOKEN.

- ref to SHA instead of branch in `RasaHQ/get-release-records-from-cms-gha`


sanchariGr (approved):
will this also be backported to all active release branches?

PR #1358, opened 2026-07-03
[INFRA-1251] Refactor Semgrep workflow

rasa-aadlv:
[JIRA](https://rasahq.atlassian.net/browse/INFRA-1251)
- Refactor Semgrep workflow to use reusable workflow from infrasec-workflows repository


rasa-aadlv:
Semgrep reusable workflow call fails: public repos can't call reusable workflows from private repos. Suggest we leave Semgrep out of this repo for now (as it was before this PR) rather than trying to work around it.

PR #1360, opened 2026-07-15
[ENG-2917] Fix ModuleNotFoundError when an actions subpackage name collides with a top-level package

djcowley:
**Proposed changes**:
- Fix a `ModuleNotFoundError` that crashed the action server on startup when custom actions live in a subpackage whose name collides with an installed top-level package — for example custom actions under `actions/agents/` together with the top-level `agents` package that `openai-agents` installs.
- `ActionExecutor._import_submodules` enumerated children with `pkgutil.walk_packages(package.__path__)` **without a prefix**. `walk_packages` imports each subpackage by its **bare leaf name** to recurse into it, so `agents` resolved to the unrelated top-level `agents` package; its submodules (`_config`, `_debug`, …) were then treated as `actions.agents.*`, and `importlib.import_module("actions.agents._config")` raised `ModuleNotFoundError`. `register_package` turns that `ImportError` into `sys.exit(1)`, so the server never starts.
- Switch to `pkgutil.iter_modules(package.__path__, prefix=package.__name__ + ".")`, which lists only the immediate children of the package path using fully-qualified names and imports nothing itself. The method's existing recursion descends correctly with those fully-qualified names, so a same-named top-level package can no longer shadow a local subpackage. The `recursive` parameter's behaviour is unchanged.

**Error observed**:
```
ModuleNotFoundError: No module named 'actions.agents._config'
```

**Test**: Added `test_submodule_name_not_shadowed_by_top_level_package` in `tests/test_executor.py`. It is self-contained (it creates its own top-level shadow package on `sys.path`, so it does not depend on `openai-agents` being installed), fails with the exact `ModuleNotFoundError` before the fix, and passes after. All existing `test_load_*` / `test_reload_*` importer tests still pass (55 passed). `make lint` and `make types` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelog/1360.bugfix.md
+++ b/changelog/1360.bugfix.md
@@ -1,0 +1,1 @@
+Custom actions placed in a subpackage whose name collides with an installed top-level package (for example `actions/agents/` alongside the `agents` package pulled in by `openai-agents`) are now imported correctly. Previously the action server failed to start with a `ModuleNotFoundError`.
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -135,6 +135,22 @@ def dispatcher() -> CollectingDispatcher:
     return CollectingDispatcher()
 
 
+@pytest.fixture()
+def shadow_package_name() -> Generator[Text, None, None]:
+    """Unique package name for collision testing.
+
+    On teardown, clears any modules it leaked into ``sys.modules`` (import
+    state is global and would pollute other tests).
+    """
+    name = "agentsshadow_" + "".join(
+        random.choice(string.ascii_lowercase) for _ in range(10)
+    )
+    yield name
+    for module_name in list(sys.modules):
+        if module_name == name or module_name.startswith(f"{name}."):
+            del sys.modules[module_name]
+
+
 def test_load_action_from_init(executor: ActionExecutor, package_path: Text):
     # Actions should be loaded from packages
     _write_action_file(package_path, "__init__.py", "InitAction", "init_action")
@@ -174,6 +190,7 @@ def test_submodule_name_not_shadowed_by_top_level_package(
     package_path: Text,
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
+    shadow_package_name: Text,
 ) -> None:
     """A subpackage must not be shadowed by a same-named top-level package.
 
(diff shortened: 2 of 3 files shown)


tests/test_executor.py
```diff
@@ -167,6 +169,65 @@ def test_load_submodules_in_namespace_package(
...
+    #    have. This mirrors the top-level ``agents`` package from openai-agents.
+    top_level = tmp_path / shadow
+    top_level.mkdir()
+    (top_level / "__init__.py").write_text("")
+    (top_level / "_config.py").write_text("")
+    monkeypatch.syspath_prepend(str(tmp_path))
+
+    # 2. The registered package contains a subpackage of the same name holding
+    #    an action, and deliberately no ``_config`` submodule.
+    subpackage_path = os.path.join(package_path, shadow)
```
ancalita, line 217: To double-check, does the `package_path` fixture correspond to an actions/ package in an agent project?

tests/test_executor.py
```diff
@@ -167,6 +169,65 @@ def test_load_submodules_in_namespace_package(
...
+        executor._import_submodules(package_name)
+
+        # register_package is the real entrypoint and must not sys.exit.
+        executor.register_package(package_name)
+    finally:
+        # monkeypatch restores sys.path automatically; drop the leaked
+        # top-level modules so they can't affect other tests.
+        for module_name in list(sys.modules):
+            if module_name == shadow or module_name.startswith(f"{shadow}."):
+                del sys.modules[module_name]
```
ancalita: can we instead have the `shadow` as an autouse pytest fixture, so this cleanup is part of the fixture code rather than of the test?

ancalita (commented):
Thanks for jumping on this so quickly 💯

changelog/1360.bugfix.md
ancalita, line 1: Also double-checking, are you backporting this to the supported release branches?

tests/test_executor.py
djcowley: Yes that correct.

tests/test_executor.py
djcowley: Yes good call will add it as a fixture, should it be autouse though since it is only used by this test?

ancalita: Ah yeah, no need to set autouse=True 👍🏻

changelog/1360.bugfix.md
djcowley: Yes will backport to 3.15.x, 3.16.x and 3.17.x  and then update the corresponding branches in Rasa pro to pin the new rasa-sdk version.

ancalita (approved):

PR #1373, opened 2026-07-30
[ENG-2854] Enable Renovate for poetry dependency updates

diana-create:
**Proposed changes**:
- Add root `renovate.json5` to onboard `rasa-sdk` to the shared Studio Renovate pattern ([ENG-2854](https://rasahq.atlassian.net/browse/ENG-2854), epic [PF-1454](https://rasahq.atlassian.net/browse/PF-1454)).
- Poetry only (`enabledManagers: ["poetry"]`): no npm/`package.json` in this repo. 3-day cooldown, Mon/Wed/Fri schedule, grouped minor/patch + security, skip majors/0.x.
- **Base branches** (`main`, `3.18.x`–`3.15.x`): chosen from rasa-sdk release lines still within the commercial 9-month End of Software Maintenance window ([release & maintenance policy](https://rasa.com/rasa-product-release-and-maintenance-policy) — associated packages). Dates come from each SDK `x.y.0` tag (which lockstep with Rasa Pro initial release dates). Drop `3.15.x` after EoSM **2026-08-26**. `3.14.x` and older are past EoSM.
- No per-repo InfraSec ticket for GitHub PAT: org-level Mend config already supplies GitHub auth for onboarded repos.
- Dependabot stays until [PF-1465](https://rasahq.atlassian.net/browse/PF-1465).
- **Also refresh `GET_PIP_SHA256` in `Dockerfile`.** The pin guards `https://bootstrap.pypa.io/get-pip.py`, which is unversioned and regenerated on pip releases. After **pip 26.2**, the old hash no longer matched, so **Build dev Docker image** failed on every PR (including this one). Updating the pin here matches how this repo usually handles dependency/bootstrap churn — as part of a maintenance batch (e.g. PF-1229 / PF-1492), not as a standalone checksum-only PR. Same stale pin remains on `3.16.x`–`3.18.x` (backports follow separately if needed). Closed the abandoned standalone attempt in #1375.

### Known limitation — Mend Silent mode
Mend Dependency Updates may remain Silent until flipped Interactive. This PR lands config only; live PRs / Dependency Dashboard are a follow-up after the Mend GitHub App has access to `RasaHQ/rasa-sdk`.

**Security scan:** Snyk Code skipped — changes are JSON5 config + a Dockerfile `ARG` pin (no first-party Python/TS/JS).

**Status (please check what you already did)**:
- [x] made PR ready for code review
- [ ] added some tests for the functionality
[...]

PR #1375, opened 2026-07-30
Fix stale get-pip.py checksum breaking Docker image builds

diana-create:
**Proposed changes**:
- Update the pinned `GET_PIP_SHA256` in `Dockerfile` to match the `get-pip.py` currently served by PyPA. The old pin (`a341e1a4…`) was invalidated by the **pip 26.2** release, which regenerated `https://bootstrap.pypa.io/get-pip.py`. Because that URL is unversioned, the checksum guard started rejecting every download and the **Build dev Docker image** job failed on every branch carrying the pin — deterministically, so re-running CI could never clear it.

This is a CI/build fix only — no `rasa_sdk` source changes. It unblocks all currently open PRs, which are blocked on this required check through no fault of their own.

**Verification** (locally, linux/arm64):
- Full image builds end to end from a clean `git archive` export, so the build context matched a CI checkout.
- Built image runs: `rasa_sdk` imports, `python -m rasa_sdk --help` prints the action-server usage, and the shipped venv reports `setuptools 82.0.1` — still in sync with `poetry.lock`, confirming the newer setuptools bundled in `get-pip.py` is correctly pinned back down.
- Base stage rebuilt with `--no-cache-filter base` to force the checksum step to actually re-execute rather than hit cache.
- `hadolint` clean, so **Lint Dockerfile** stays green.
- CI: **Build dev Docker image** passed on this PR after the pin update (~6 min full multi-platform build; previously failed at ~53s).

**Note:** the same stale pin is also on `3.16.x`, `3.17.x` and `3.18.x`, which will hit this on their next PR. `3.15.x` is unaffected (no such pin). Backports to follow separately.

**Follow-up worth considering:** any hardcoded hash of an unversioned URL will go stale again on the next pip release, and Renovate cannot manage a hand-rolled `ARG`. A durable fix is to drop the `get-pip.py` bootstrap in favour of `ensurepip`/`python3-pip`, but that needs care around the `update-alternatives` path that assumes get-pip's `/usr/local/bin/pip3` location — too invasive to bundle with an urgent unblock.


diana-create:
Closing in favour of folding the hash refresh into #1373 (dependency/tooling PR), which matches how this repo usually handles bootstrap/dependency pin churn rather than a dedicated checksum-only PR.

No separate backports were opened from this branch.

